PR #27, opened 2021-11-15
add config to support remote endpoint test

### LINKED ISSUE #8: Add support to run tests against specific endpoint
Current repo is only running against localhost:5601. Extend this so that it can run against an input endpoint

Tengda-He:
Signed-off-by: Tengda He <tengh@amazon.com>

### Description
- adding a separate cypress environment variable for OpenSearch endpoint url, the baseUrl will be reserved for OpenSearchDashboard endpoint. This way allows easy config for API call to OpenSearch endpoint.
- move the username and password parameter to  cypress environment variable
- cypress environment variable supports CLI overrides as well.

### Test
- all existing test suites passed for remote endpoint in staging environment
- fgac tests passed
- non fgac tests passed

### Issues Resolved
#8 
 
### Check List

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -4,5 +4,10 @@
   "reporter": "cypress-multi-reporters",
   "reporterOptions": {
     "configFile": "reporter-config.json"
+  },
+  "env": {
+    "openSearchUrl": "http://localhost:9200",
+    "userName": "admin",
+    "password": "admin"
   }
 }
--- a/cypress/integration/security-disabled/dashboard_filtering_spec.js
+++ b/cypress/integration/security-disabled/dashboard_filtering_spec.js
@@ -12,7 +12,7 @@ import { TestFixtureHandler, CommonUI, DashboardPage, MiscUtils } from '@opensea
  * 8) Test adding another pie graph to the dashboard and applying a filter to both graphs
  */
 
-const testFixtureHandler = new TestFixtureHandler(cy)
+const testFixtureHandler = new TestFixtureHandler(cy, Cypress.env('openSearchUrl'))
 const commonUI = new CommonUI(cy)
 const dashboardPage = new DashboardPage(cy)
 const miscUtils = new MiscUtils(cy)
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,6 @@
 node_modules
 cypress/videos/
+cypress/results/
 .DS_Store
 **/.DS_Store
+
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,4 +1,6 @@
 - [Contributing to OpenSearch](#contributing-to-opensearch)
+- [Developer Guide](#developer-guide)
+  - [Run Tests](#run-tests)
 - [First Things First](#first-things-first)
 - [Ways to Contribute](#ways-to-contribute)
   - [Bug Reports](#bug-reports)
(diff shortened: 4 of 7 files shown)


cypress.json
```diff
@@ -4,5 +4,11 @@
   "reporter": "cypress-multi-reporters",
   "reporterOptions": {
     "configFile": "reporter-config.json"
+  },
+  "env": {
+    "openSearchUrl": "http://localhost",
```
seraphjiang, line 9: how about dashboards url?
e.g.
es endpoint is: https://foo.com/
osd endpoint is https://foo.com/_dashbaords

tianleh:
Can you share the CLIs you used for testing?

cypress.json
```diff
@@ -4,5 +4,11 @@
...
+    "openSearchPort": 9200,
```
seraphjiang, line 10: we need to change cypress.json file in order to pick up the test settings. could we make it part of system environment variable or input of launch script parameter, so we don't need to change file for different environment

cypress.json
Tengda-He: we already use the reserved word "baseUrl": "http://localhost:5601" in cypress.json for dashboards url, when we call the cy.visit() api, the baseUrl is added as prefix.

cypress.json
Tengda-He: the cypress environment variable can be override by CLI at execution time, ref: https://docs.cypress.io/guides/references/configuration#Environment-Variables

Tengda-He:
@tianleh here is the CLI for test:
npx cypress run --spec "cypress/integration/security-disabled/*.js" --env "openSearchUrl=https://search-tengh-staging-osd-11-x64-01-l7zlsfsvhn2wgtzqns2322zy5q.us-west-2.es-staging.amazonaws.com,openSearchPort=443" --config "baseUrl=https://search-tengh-staging-osd-11-x64-01-l7zlsfsvhn2wgtzqns2322zy5q.us-west-2.es-staging.amazonaws.com/_dashboards"

cypress.json
tianleh: The style for os and osd is inconsistent. 1. the variable name. 2. whether port is needed.

Probably use this opportunity to unify the experience. 

```
openSearchUrl
openSearchDashboardsUrl
```

and get rid of the ports. This may require some more changes inside https://github.com/opensearch-project/opensearch-dashboards-test-library/blob/bc5202067ddf73e6bfd98e475e02500647cb1b73/common-utils/network/test-fixture-handler.js#L3

Tengda-He: For question 1 which to use openSearchDashboardsUrl or baseUrl for variable name:

baseUrl is a reserved word integrated with Cypress framework, quote "URL used as prefix for cy.visit() or cy.request() command's URL". If we decide to create a new env variable like openSearchDashboardsUrl, it loses that prefix function. And we need to rewrite lots of existing test cases, as they are written in cy.visit(relative_path), which implies visiting baseUrl + relative_path. Which I don't suggest to change.

For question 2 whether port is needed:
I think it is necessary to have port configurable. The reason is that local testing, to distinguish frontend and backend, we use port 5601 and 9200. But for remote testing to distinguish frontend and backend, we use relative path like */_dashboard for frontend and */ for backend, and the port is the same (80 for http and 443 for https). So to support both local and remote test, I suggest to keep port configuration. Another tradeoff to note, if we use the baseUrl parameter then the port must be embedded in it. like "baseUrl": "http://localhost:5601", we can't set the port variable separately.

open to more suggestions.

tianleh: For Q2, what I mean is that we can embed port into the endpoint for open search, similar to baseUrl.

Tengda-He: got it, from test-fixture-handler class, it uses like ${this.hostname}:${this.port} only, we definitely can combine these two variable to one.

cypress/integration/security-disabled/dashboard_filtering_spec.js
```diff
@@ -205,7 +205,7 @@ describe('dashboard filtering', () => {
         // Inconsistency: original code says that the goal label should have "7,544",
         // but sometimes the goal displays "7,565". It may have been related to a
         // data loading issue.
-        commonUI.checkValuesExistInComponent('svg > g > g > text.chart-label', ['7,544', '39.958%'])
+        // commonUI.checkValuesExistInComponent('svg > g > g > text.chart-label', ['7,544', '39.958%'])
```
seraphjiang, line 208: why comment out thsi?

seraphjiang (commented):
it is nice to support remote endpoint with config file. but i would like to see test could load parameters from environment variable. so we don't need to change source code(config file) for different runtime value.

seraphjiang (approved):

cypress/integration/security-disabled/dashboard_filtering_spec.js
Tengda-He: This one is a little flaky, it failed on local machine

tianleh:
> it is nice to support remote endpoint with config file. but i would like to see test could load parameters from environment variable. so we don't need to change source code(config file) for different runtime value.

The values in config file are just default placeholders. From the CLI @Tengda-He shared, it is already dynamic.

tianleh:
@Tengda-He Can you check the failed checks?

tianleh:
Add to contributing doc about the usage of this feature.

Tengda-He:
add the developer guide section

tianleh (approved):
LGTM.

Just a small PR suggestion to have all your commits so that reviewers can easily see the diff between each iteration. We can always use the "Squash and Merge" option to consolidate all commits to one during merge.

PR #64, opened 2021-12-15
Onboard initial AD Dashboard plugin cypress tests

### LINKED ISSUE #38: onboard AD plugin

### LINKED ISSUE #63: Change strategy for toggling security
Currently the strategy for testing security enabled or disabled is to have different specs under `with-security` and `without-security` dirs for each component, where `with-security` is essentially the same spec, but has an added first step which involves logging in to the cluster from the login page, using username/password env vars.

A simpler soln can be overriding the `cy.visit()` cmd to pass in a username/password directly, based on the value of some `SECURITY_ENABLED` env var.  This allows you to write a single spec, and be able to run the same set of tests as security enabled or disabled by only toggling the env var as `true` or `false`. Because most tests won't involve any direct UI changes based on the enablement of security or not, this makes it easier to reuse code in both scenarios.

This is the current way that individual plugins are handling cypress tests as well, so by using this strategy it will be less work for plugins to onboard tests without having to refactor into security-enabled and security-disabled tests.

ohltyler:

### Description
This PR adds some initial tests for the AD Dashboards plugin's dashboard page, ported over from [here](https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/blob/main/.cypress/integration/ad/dashboard/ad_dashboard.spec.ts). Created some files that future plugins can add to for their plugin needs (constants can go in `utils/constants`, custom commands can go in `support/commands`).

Other changes include:
- Refactoring how tests are ran based on security being enabled or disabled. Sets up a `SECURITY_ENABLED` cypress env var which can be toggled to run the same set of tests for both scenarios. More details on the change in #63.
- Refactoring the core dashboards tests under a base `core-opensearch-dashboards` dir to easily differentiate between plugin tests.
- Adding a workflow using bundled artifacts to test all plugin tests (which is currently just AD)

NOTE: minor formatting changes (in workflows, for example) can be ignored. Will have a follow-up PR to integrate the codebase with `prettier` which will do a full reformat over all `.js` files.
 
### Issues Resolved
Closes #63 
Closes #38 

 
### Check List


cypress.json
```diff
@@ -9,7 +9,7 @@
   "viewportHeight": 1320,
   "env": {
     "openSearchUrl": "http://localhost:9200",
-    "userName": "admin",
-    "password": "admin"
-  }
+    "SECURITY_ENABLED": false
+  },
+  "experimentalNetworkStubbing": true
```
ohltyler, line 14: This is needed to use the `cy.route2` cmd which is necessary for calls to the AD server.

ohltyler:
CI seems to keep failing with this error - any idea @tianleh?
```
  1) dashboard filtering
       "before all" hook for "Buffer test for importing test environment data":
     CypressError: `cy.request()` failed trying to load:

http://localhost:9200/animals-dogs-2018-01-01

We attempted to make an http request to this URL but the request failed without a response.

We received this error at the network level:

  > Error: socket hang up
```

ohltyler:
Setting to draft to set up plugin workflows first.

seraphjiang:
Is it connection timeout. What's size of the test data

> CI seems to keep failing with this error - any idea @tianleh?
> 
> ```
> 
>   1) dashboard filtering
> 
>        "before all" hook for "Buffer test for importing test environment data":
> 
>      CypressError: `cy.request()` failed trying to load:
> 
> 
> 
> http://localhost:9200/animals-dogs-2018-01-01
> 
> 
> 
> We attempted to make an http request to this URL but the request failed without a response.
> 
> 
> 
> We received this error at the network level:
> 
> 
> 
>   > Error: socket hang up
> 
> ```

ohltyler:
> Is it connection timeout. What's size of the test data
> 
> > CI seems to keep failing with this error - any idea @tianleh?
> > ```
> > 
> >   1) dashboard filtering
> > 
> >        "before all" hook for "Buffer test for importing test environment data":
> > 
> >      CypressError: `cy.request()` failed trying to load:
> > 
> > 
> > 
> > http://localhost:9200/animals-dogs-2018-01-01
> > 
> > 
> > 
> > We attempted to make an http request to this URL but the request failed without a response.
> > 
> > 
> > 
> > We received this error at the network level:
> > 
> > 
> > 
> >   > Error: socket hang up
> > ```

I found out that after refactoring, I changed the functionality and was running the vanilla tests in the bundled workflow. I've since removed that and all tests are passing now.

tianleh:
Huge progress! 

Just want to mention this location may need to be updated as well (similar to the change in Github action). https://github.com/opensearch-project/opensearch-dashboards-functional-test/blob/main/integtest.sh#L82

This is how infra team will run test cases inside this repo.

.github/workflows/cypress-workflow-vanilla-snapshot-based.yml
```diff
@@ -70,8 +70,8 @@ jobs:
         uses: cypress-io/github-action@v2
         with:
           working-directory: monetery-test
-          command: yarn run cypress run --browser chrome --headless --spec 'cypress/integration/vanilla-opensearch-dashboards/*.js'
-          wait-on: 'http://localhost:5601'
+          command: yarn run cypress run --browser chrome --headless --env SECURITY_ENABLED=true --spec 'cypress/integration/core-opensearch-dashboards/vanilla-opensearch-dashboards/*.js'
```
tianleh, line 73: Line 34 indicates that this is a min artifact. Shall this be `SECURITY_ENABLED=false`?

.github/workflows/cypress-workflow-vanilla-source-based.yml
```diff
@@ -88,8 +88,8 @@ jobs:
         uses: cypress-io/github-action@v2
         with:
           working-directory: monetery-test
-          command: yarn run cypress run --browser chrome --headless --spec 'cypress/integration/vanilla-opensearch-dashboards/*.js'
-          wait-on: 'http://localhost:5601'
+          command: yarn run cypress run --browser chrome --headless --env SECURITY_ENABLED=true --spec 'cypress/integration/core-opensearch-dashboards/vanilla-opensearch-dashboards/*.js'
```
tianleh, line 91: There is no plugin for this flow. Shall this be `false` instead? Interesting that specifying `true` didn't fail the test.

.github/workflows/cypress-workflow-vanilla-snapshot-based.yml
ohltyler: Yes good catch, I'll explicitly change to false.

.github/workflows/cypress-workflow-vanilla-source-based.yml
ohltyler: Yes, should be false - I'll update here as well.

ohltyler: I think it still succeeds since the only difference is the requests having the added `admin/admin` auth, which is probably just ignored if no security plugin exists.

ohltyler:
> 

Good catch - I've updated `integtest.sh` and added a line to test the plugin test suite. Also explicitly set security to `false` in the vanilla workflows.

tianleh (approved):

Tengda-He (approved):

tianleh:
Once this PR is merged, there will be two follow up items.

1) for @tianleh , pull in latest change and run the integtest.sh from infra's repo https://github.com/opensearch-project/opensearch-dashboards-functional-test/issues/67

2) for @ohltyler check if AD's existing CI (Github action for Cypress) can check out this repo and run the AD test cases inside this repo. This is to avoid duplicated test cases. https://github.com/opensearch-project/opensearch-dashboards-functional-test/issues/65 Let us know if you could take this task.

ohltyler:
> Once this PR is merged, there will be two follow up items.
> 
> 1. for @tianleh , pull in latest change and run the integtest.sh from infra's repo [[FEATURE] Verify https://github.com/opensearch-project/opensearch-dashboards-functional-test/pull/64 from infra #67](https://github.com/opensearch-project/opensearch-dashboards-functional-test/issues/67)
> 2. for @ohltyler check if AD's existing CI (Github action for Cypress) can check out this repo and run the AD test cases inside this repo. This is to avoid duplicated test cases. [[FEATURE] call AD test cases inside this repo from AD's CI process #65](https://github.com/opensearch-project/opensearch-dashboards-functional-test/issues/65) Let us know if you could take this task.

@tianleh can you explain the purpose of running tests added here, inside of the individual plugin CI? From my understanding, the tests added in this repo should be a subset / more high-level integ tests, while the individual plugin test suite can have longer, more detailed tests.

Is the long-term plan to move most integ tests here, or are you suggesting to add a CI workflow in the plugin that runs these tests, as sort of an extra layer of sanity tests at the individual plugin level?

tianleh:
> @tianleh can you explain the purpose of running tests added here, inside of the individual plugin CI? From my understanding, the tests added in this repo should be a subset / more high-level integ tests, while the individual plugin test suite can have longer, more detailed tests.
> 
> Is the long-term plan to move most integ tests here, or are you suggesting to add a CI workflow in the plugin that runs these tests, as sort of an extra layer of sanity tests at the individual plugin level?

@ohltyler The long term plan is to move all integ tests here in this repo (we may need to add some folder structure to distinguish release signoff test vs regular integ tests though). However, it is not needed in this iteration. 

The immediate need is to "add a CI workflow in the plugin that runs these tests, as sort of an extra layer of sanity tests at the individual plugin level". It can serve as two purposes, one to plugin level sanity test, second to proof the technical feasibility to have plugin repo able to checkout this repo and run test cases inside this repo.

cc @seraphjiang @Tengda-He for visibility

ohltyler:
> > @tianleh can you explain the purpose of running tests added here, inside of the individual plugin CI? From my understanding, the tests added in this repo should be a subset / more high-level integ tests, while the individual plugin test suite can have longer, more detailed tests.
> > Is the long-term plan to move most integ tests here, or are you suggesting to add a CI workflow in the plugin that runs these tests, as sort of an extra layer of sanity tests at the individual plugin level?
> 
> @ohltyler The long term plan is to move all integ tests here in this repo (we may need to add some folder structure to distinguish release signoff test vs regular integ tests though). However, it is not needed in this iteration.
> 
> The immediate need is to "add a CI workflow in the plugin that runs these tests, as sort of an extra layer of sanity tests at the individual plugin level". It can serve as two purposes, one to plugin level sanity test, second to proof the technical feasibility to have plugin repo able to checkout this repo and run test cases inside this repo.
> 
> cc @seraphjiang @Tengda-He for visibility

Got it, thanks for the clarification. That makes sense. I'll get started on that in the AD plugin.

PR #94, opened 2022-02-15
Onboard index management cypress test

### LINKED ISSUE #73: index-management-dashboards-plugin
See the main tracking issue for instructions https://github.com/opensearch-project/opensearch-dashboards-functional-test/issues/72

bowenlan-amzn:
Signed-off-by: bowenlan-amzn <bowenlan23@gmail.com>

### Description

Apart from the copy/paste cypress test over from IM dashboards repo, this PR investigates running cypress against an AWS OpenSearch domain with fgac enabled.

AWS OpenSearch domain has limitations that affect existing tests and future test writing:

- Some cluster/index settings are not exposed to user, so user cannot read or change them. 
- Even admin user cannot do read, update or delete operation to the system index. 

### Issues Resolved

#73 

### Check List


cypress/utils/constants.js
```diff
@@ -50,3 +50,32 @@ export const DETECTOR_STATE_INIT = 'Initializing';
 export const DETECTOR_STATE_RUNNING = 'Running';
 
 export const DETECTOR_STATE_FEATURE_REQUIRED = 'Feature required';
+
+/**
+ *****************************
+ IM PLUGIN CONSTANTS
```
tianleh, line 56: looks like this file will grow larger than expected. Probably you can create a separate file for ISM constants. 

cc @ohltyler

cypress/support/commands.js
```diff
@@ -147,3 +163,91 @@ Cypress.Commands.add(
     cy.wait('@createDetector');
   }
 );
+
+/**
+ *****************************
+ IM PLUGIN COMMANDS
```
tianleh, line 169: Same here cc @ohltyler

tianleh:
Great progress! Will take a closer look once PR is ready for review.

cypress/support/commands.js
ohltyler: Agreed. I think this can be done in a similar strategy to my [first comment](https://github.com/opensearch-project/opensearch-dashboards-functional-test/pull/94#discussion_r809456457). 

To get added in the cypress config properly, be sure to import your new commands file in [cypress/support/commands/index.js](https://github.com/opensearch-project/opensearch-dashboards-functional-test/blob/main/cypress/support/index.js#L17) (ref: https://github.com/opensearch-project/opensearch-dashboards-functional-test/pull/94#discussion_r809456457)

ohltyler:
Regarding the movement of the existing AD constants / custom commands: I will handle in a follow-up PR. @bowenlan-amzn I recommend moving your added coded to IM-specific files, and you can leave the AD ones for now.

cypress/utils/constants.js
tianleh: I think it makes sense. 

> Agreed that this may grow too large and can lead to annoying merge conflicts. I think a good solution may be moving these sorts of files (`constants.js`, `helpers.js`, etc.) within the plugin directory (e.g., `plugins/index-management-dashboards-plugin/constants.js`.
> 
> @bowenlan-amzn @tianleh what do you think?

LGTM.

bowenlan-amzn: Thanks, I will follow this way.

cypress/utils/plugins/anomaly-detection-dashboards-plugin/commands.js
```diff
@@ -0,0 +1,94 @@
+/*
```
ohltyler, line 1: What about putting all of these files (`commands.js`, `constants.js`, `helpers.js`, etc.) in a `utils` dir per plugin (`cypress/plugins/anomaly-detection-dashboards-plugin/utils/*`)? 

To me, it seems these should be bundled in one location. Lmk what you think.

cypress/utils/constants.js
```diff
@@ -4,49 +4,7 @@
...
-// TODO: when repo is onboarded to typescript we can convert these detector states into an enum
-export const DETECTOR_STATE_DISABLED = 'Stopped';
-
-export const DETECTOR_STATE_INIT = 'Initializing';
-
-export const DETECTOR_STATE_RUNNING = 'Running';
-
-export const DETECTOR_STATE_FEATURE_REQUIRED = 'Feature required';
+export * from './plugins/anomaly-detection-dashboards-plugin/constants';
+export * from './plugins/index-management-dashboards-plugin/constants';
```
ohltyler, line 10: Any specific reason for exporting here? Seems we could just export and import at the plugin-dir level to simplify things.

cypress/support/index.js
```diff
@@ -27,6 +29,15 @@ Cypress.on('uncaught:exception', (err) => {
   }
 });
 
-if (Cypress.env('SECURITY_ENABLED')) {
-  Cypress.env('openSearchUrl', 'https://localhost:9200');
+// Proxy layer of AWS OpenSearch domain redirects to login page
+//  if you haven't authenticate
+if (Cypress.env('AWS_DOMAIN')) {
```
ohltyler, line 34: What about an opensource domain with security enabled? I don't think we should remove the existing logic.

package.json
```diff
@@ -7,7 +7,7 @@
     "test": "echo \"Error: no test specified\" && exit 1",
     "cypress:open": "cypress open",
     "cypress:run-without-security": "env TZ=America/Los_Angeles cypress run --browser chrome --headless --env SECURITY_ENABLED=false",
-    "cypress:run-with-security": "env TZ=America/Los_Angeles cypress run --browser chrome --headless --env SECURITY_ENABLED=true",
+    "cypress:run-with-security": "env TZ=America/Los_Angeles cypress run --browser chrome --headless --env SECURITY_ENABLED=true,openSearchUrl=https://localhost:9200",
```
ohltyler, line 10: Similar to previous comment, if we keep the existing logic we don't need to add this.

CONTRIBUTING.md
```diff
@@ -68,12 +70,16 @@ These tests run in headless mode by default. You can also manually trigger the t
 $ yarn cypress open
 ```
 
-And you can override certain cypress config or environment variable by appling additional cli arguments, for example to override the baseUrl and OpensearchUrl to test a remote endpoint:
+And you can override certain [cypress config or environment variable](cypress.json) by applying additional cli arguments, for example to override the baseUrl and OpensearchUrl to test a remote **AWS OpenSearch endpoint**:
```
ohltyler, line 73: nit: `OpensearchUrl` -> `openSearchUrl`

cypress/utils/constants.js
bowenlan-amzn: Through this way, tests look cleaner to me, all constants import from single place. 
I don't think import contants from plugin constant file is what we want, we still want a single source, but just refactor those to not explode the main constants.js.

cypress/support/index.js
bowenlan-amzn: `SECURITY_ENABLED` is not supposed to tuning the url, because we provide url directly through `openSearchUrl`

cypress/utils/plugins/anomaly-detection-dashboards-plugin/commands.js
ohltyler: If so, can we move all `commands/js` / `constants.js` / `helpers.js` files to a single directory so it's not split up?

cypress/utils/constants.js
ohltyler: It is very common to have constants only scoped for their specific modules / components within a project - this seems to add unnecessary overhead of importing a constant from a central constants file which is exporting it from some local place to begin with. Additionally, it adds some confusing loading issues as mentioned [here](https://github.com/opensearch-project/opensearch-dashboards-functional-test/pull/94#discussion_r810173438), which seems like a common use case, where global constants may want to be used by local / plugin-level constants. 

The tests would still look clean, as they can import from `./constants` or `./utils/constants`, instead of `../../constants`

@tianleh thoughts on this?

cypress/utils/plugins/anomaly-detection-dashboards-plugin/constants.js
```diff
@@ -0,0 +1,50 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+// since this file is export in main constant file
+// this will be loaded before main constant file
+// so cannot import BASE_PATH but have to define it directly
+const BASE_PATH = Cypress.config('baseUrl');
```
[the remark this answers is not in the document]
bowenlan-amzn: Discussed with Tyler offline, will see if we can define this as static so that still can import.

cypress/utils/plugins/anomaly-detection-dashboards-plugin/constants.js
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { BASE_PATH } from '../../base_constants';
```
ohltyler, line 6: Can you update for [gantt charts](https://github.com/opensearch-project/opensearch-dashboards-functional-test/blob/main/cypress/integration/plugins/gantt-chart-dashboards/gantt_ui.spec.js#L8) as well?

cypress/utils/plugins/index-management-dashboards-plugin/commands.js
```diff
@@ -0,0 +1,160 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { BACKEND_BASE, IM_API, IM_CONFIG_INDEX } from '../../constants';
+
+Cypress.Commands.add('deleteAllIndices', () => {
```
ohltyler, line 8: Seems this could be a general purpose command based on the name - could some of the logic be refactored into a common command and referenced in here, and maybe rename this command to something ISM-specific? At least the renaming could prevent conflicts in the future.

cypress/utils/base_constants.js
```diff
@@ -0,0 +1,7 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export const BASE_PATH = Cypress.config('baseUrl');
+export const BACKEND_BASE = Cypress.env('openSearchUrl');
```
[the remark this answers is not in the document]
bowenlan-amzn: updated

cypress/utils/plugins/anomaly-detection-dashboards-plugin/constants.js
bowenlan-amzn: So outside the utils folder, constants should all be import from `'../../../utils/constants'`

cypress/utils/plugins/index-management-dashboards-plugin/commands.js
bowenlan-amzn: Refactor all the general command that could be used by other plugin to the main command file

cypress/utils/plugins/anomaly-detection-dashboards-plugin/commands.js
bowenlan-amzn: All move to `utils`

cypress/utils/plugins/anomaly-detection-dashboards-plugin/constants.js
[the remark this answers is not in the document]
bowenlan-amzn: introduce `base_contants.js` file, use `contants.js` as a barrel for all constants

cypress/utils/constants.js
tianleh: I would prefer each test file importing their own constant file. The `cypress/utils/constants.js` ideally shall only contain common ones.

tianleh: After checking further with @bowenlan-amzn , it is because ISM tests need both the common constants and ISM specific ones. Thus they will need to always write two lines of import in their test files which is indeed annoying. Thus doing in current way could just use one line of import. I am ok with this.

CONTRIBUTING.md
```diff
@@ -68,12 +70,16 @@ These tests run in headless mode by default. You can also manually trigger the t
...
 
 ```
-$ yarn cypress run --spec "cypress/integration/core-opensearch-dashboards/vanilla-opensearch-dashboards/*.js" --env "openSearchUrl=https://foo.com" --config "baseUrl=https://foo.com/_dashboards"
+$ yarn cypress run --spec "cypress/integration/core-opensearch-dashboards/vanilla-opensearch-dashboards/*.js" --config "baseUrl=https://<endpoint>/_dashboards" --env "openSearchUrl=https://<endpoint>,SECURITY_ENABLED=true,username=admin,password=xxxxxxxx,AWS_DOMAIN=true"
 ```
 
+`SECURITY_ENABLED`: if true, the `username` and `password` passing in are used as basic authentication credentials during `cy.visit` and `cy.request`. Also please notice security enabled domain normally uses https protocol, so you may want to pass in different urls.
+
+`AWS_DOMAIN`: AWS OpenSearch domain is wrapped with proxy that redirects the visiting url to the login url. Even with auth option provided in `cy.visit`, redirection to the login url happens. So a login request before tests and cache the security cookie are needed and can be switched on by this argument.
```
tianleh, line 81: Could we be generic about the name, to be `ENDPOINT_WITH_PROXY`? As open source repo, we would like to be generic. Could you also update the related text to be generic description as well?

bowenlan-amzn: Removed the usage of aws, domain.

tianleh (approved):

seraphjiang (approved):

PR #136, opened 2022-03-17
Upgrade to Cypress 9.5.2

tianleh:
### Description

Since the functional tests have its own execution env, we have the flexibility to upgrade Cypress to more newer versions with more bug fixes and advanced features. Pick the latest one https://docs.cypress.io/guides/references/changelog#9-5-2 The current code base has Cypress 5.6.1 which is 4 major versions behind.

### Issues Resolved

Resolves https://github.com/opensearch-project/opensearch-dashboards-functional-test/issues/81

### Check List

- [x] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #144, opened 2022-03-23
Clean up flaky failures in AD plugin

ohltyler:

### Description

The logic to delete the detector in the create detector workflow test would occasionally fail, due to trying to delete it too soon after creation. Because of this failure to delete, it had cascading effects on other tests, where they were expecting empty pages / no detectors, but instead was displaying this test detector which failed to delete.

This fixes the logic in the deletion by utilizing the UI instead of making a direct API call, which provides the benefits of (1) more closely replicating how a user would interact with the plugin and (2) provides more time for the detector to be created, as the modal to delete won't be functional until the detector has been created and is available to delete.

Testing:
- Debugged failures by replicating in github runners and downloading the videos to confirm why they were failing
- Confirmed I could replicate all of the failures seen before in AD's remote integ test workflow
- Changed the AD remote integ test workflow (in a test branch) to point to this forked version of FTRepo to confirm the fix. Ran multiple times and cannot see any failures now (you can see all of the successful runs [here](https://github.com/ohltyler/anomaly-detection-dashboards-plugin-1/actions/runs/2031254251/attempts/1) by clicking through each of the attempts)

### Issues Resolved

Closes https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/issues/211

### Check List

### DIFF
--- a/cypress/integration/plugins/anomaly-detection-dashboards-plugin/create_detector_spec.js
+++ b/cypress/integration/plugins/anomaly-detection-dashboards-plugin/create_detector_spec.js
@@ -80,14 +80,10 @@ context('Create detector workflow', () => {
     cy.getElementByTestId('featureTable').contains(TEST_FEATURE_NAME);
     cy.getElementByTestId('createDetectorButton').click();
 
-    // Clean up the created detector. Extract detector ID from the detector configuration page
-    cy.getElementByTestId('detectorIdCell').within(() => {
-      cy.get('.euiText--medium')
-        .invoke('text')
-        .then((detectorId) => {
-          cy.log('Deleting detector with ID: ' + detectorId);
-          cy.deleteDetector(detectorId);
-        });
-    });
+    // Lands on the config page by default. Delete the detector to clean up.
+    cy.getElementByTestId('actionsButton').click();
+    cy.getElementByTestId('deleteDetectorItem').click();
+    cy.getElementByTestId('typeDeleteField').type('delete', { force: true });
+    cy.getElementByTestId('confirmButton').click();
   });
 });


kavilla (approved):

ohltyler:
I'm going to run the [workflow](https://github.com/ohltyler/anomaly-detection-dashboards-plugin-1/actions/runs/2031254251) a few more times to confirm there's no failures before merging.

cypress/integration/plugins/anomaly-detection-dashboards-plugin/create_detector_spec.js
```diff
@@ -80,14 +80,10 @@ context('Create detector workflow', () => {
...
-      cy.get('.euiText--medium')
-        .invoke('text')
-        .then((detectorId) => {
-          cy.log('Deleting detector with ID: ' + detectorId);
-          cy.deleteDetector(detectorId);
-        });
-    });
+    // Lands on the config page by default. Delete the detector to clean up.
+    cy.getElementByTestId('actionsButton').click();
+    cy.getElementByTestId('deleteDetectorItem').click();
```
amitgalitz, line 85: Will it wait for delete detector button to show up or basically waits for it to load the config page before trying these clicks

kavilla: It should retry by default.

ohltyler: The page and all of the components, including the dropdown with the delete option, will need to load and be functional before it can execute all of this. In other words, if following this sequence of events fails to delete the detector, then that is a separate UI bug that would need to be addressed, and wouldn't be an issue with race conditions in the test

amitgalitz (approved):

tianleh:
Great work! Just want to mention that I find another flaky one in the Cypress upgrade PR https://github.com/opensearch-project/opensearch-dashboards-functional-test/pull/136

It is `detector_list_spec.js`
https://github.com/opensearch-project/opensearch-dashboards-functional-test/runs/5649728704?check_suite_focus=true

This can be tracked and monitored separately. Feel free to merge as incremental improvement.

tianleh (approved):

ohltyler:
> Great work! Just want to mention that I find another flaky one in the Cypress upgrade PR #136
> 
> It is `detector_list_spec.js` https://github.com/opensearch-project/opensearch-dashboards-functional-test/runs/5649728704?check_suite_focus=true
> 
> This can be tracked and monitored separately. Feel free to merge as incremental improvement.

Thanks for bringing that up - I've been trying to reproduce that one with no luck, may be related to the upgrade, as I've never seen that in 5.6 cypress. Will consider it a separate issue.

Issue #184, opened 2022-04-14
[FEATURE] Onboard notifications sanity tests


kavilla:
New plugin

https://github.com/opensearch-project/notifications/tree/main/dashboards-notifications

tianleh:
Shall we add `2.0.0` label for this issue?

PR #185, opened 2022-04-14
[Notifications] onboard sanity tests for notifications plugin

kavilla:
### Description

New plugin. Just copy and pasted the existing tests from this repo:

https://github.com/opensearch-project/notifications/tree/main/dashboards-notifications

I am not positive about the flakiness this introduces but will handle it in subsequent tests.

### Issues Resolved

https://github.com/opensearch-project/opensearch-dashboards-functional-test/issues/184

### Check List

### DIFF
--- a/.github/workflows/notifications-release-e2e-workflow.yml
+++ b/.github/workflows/notifications-release-e2e-workflow.yml
@@ -1,0 +1,13 @@
+name: Notifications Release tests workflow in Bundled OpenSearch Dashboards
+on:
+  pull_request:
+    branches:
+      - main
+      - dev-*
+jobs:
+  tests:
+    uses: ./.github/workflows/release-e2e-workflow-template.yml
+    with:
+      test-name: Notifications
+      test-command: yarn cypress:run-with-security --browser chromium --spec 'cypress/integration/plugins/notifications-dashboards/*'
+      version: 2.0.0-rc1
--- a/cypress/integration/plugins/notifications-dashboards/channels.spec.js
+++ b/cypress/integration/plugins/notifications-dashboards/channels.spec.js
@@ -1,0 +1,275 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/// <reference types="cypress" />
+
+import { delay, NOTIFICATIONS_PLUGIN_NAME } from '../../../utils/plugins/notifications-dashboards/constants';
+import { BASE_PATH } from "../../../utils/base_constants";
+
+describe('Test create channels', () => {
+  beforeEach(() => {
+    cy.visit(`${BASE_PATH}/app/${NOTIFICATIONS_PLUGIN_NAME}#create-channel`);
+    cy.wait(delay * 3);
+  });
+
+  it('creates a slack channel and send test message', () => {
+    cy.get('[data-test-subj="create-channel-create-button"]').click();
+    cy.contains('Some fields are invalid.').should('exist');
+
+    cy.get('[placeholder="Enter channel name"]').type('Test slack channel');
+    cy.get('.euiCheckbox__input[id="alerting"]').click({ force: true });
+    cy.get('[data-test-subj="create-channel-slack-webhook-input"]').type(
+      'https://sample-slack-webhook'
+    );
+    cy.wait(delay);
+    cy.get('[data-test-subj="create-channel-send-test-message-button"]').click({
+      force: true,
+    });
+    cy.wait(delay);
+    // This needs some time to appear as it will wait for backend call to timeout
+    cy.contains('test message.').should('exist');
+
+    cy.get('[data-test-subj="create-channel-create-button"]').click({
+      force: true,
+    });
+    cy.contains('successfully created.').should('exist');
+  });
+
+  it('creates a chime channel', () => {
+    cy.get('[placeholder="Enter channel name"]').type('Test chime channel');
+    cy.get('.euiCheckbox__input[id="reports"]').click({ force: true });
+
+    cy.get('.euiSuperSelectControl').contains('Slack').click({ force: true });
+    cy.wait(delay);
+    cy.get('.euiContextMenuItem__text')
+      .contains('Chime')
+      .click({ force: true });
+    cy.wait(delay);
+
+    cy.get('[data-test-subj="create-channel-chime-webhook-input"]').type(
+      'https://sample-chime-webhook'
+    );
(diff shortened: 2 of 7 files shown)


kavilla:
@qreshi

tianleh:
Can you add a PR check workflow for this plugin similar to other plugins?

kavilla:
> Can you add a PR check workflow for this plugin similar to other plugins?

I think so! Do you manually create that from the workflow yaml file? The only thing is does it require an existing run? I'm not positive if there has been a successful run with OpenSearch that includes the notifications plugin.

tianleh:
> I think so! Do you manually create that from the workflow yaml file? The only thing is does it require an existing run? I'm not positive if there has been a successful run with OpenSearch that includes the notifications plugin.

I created a template for such. Check this example. https://github.com/opensearch-project/opensearch-dashboards-functional-test/blob/main/.github/workflows/gantt-chart-release-e2e-workflow.yml

If it cannot pass at this moment, it is fine to merge since at least it has the set up.

kavilla:
> > I think so! Do you manually create that from the workflow yaml file? The only thing is does it require an existing run? I'm not positive if there has been a successful run with OpenSearch that includes the notifications plugin.
> 
> I created a template for such. Check this example. https://github.com/opensearch-project/opensearch-dashboards-functional-test/blob/main/.github/workflows/gantt-chart-release-e2e-workflow.yml
> 
> If it cannot pass at this moment, it is fine to merge since at least it has the set up.

Oh ha my `main` branch was pointed to my fork's main

kavilla:
> > I think so! Do you manually create that from the workflow yaml file? The only thing is does it require an existing run? I'm not positive if there has been a successful run with OpenSearch that includes the notifications plugin.
> 
> I created a template for such. Check this example. https://github.com/opensearch-project/opensearch-dashboards-functional-test/blob/main/.github/workflows/gantt-chart-release-e2e-workflow.yml
> 
> If it cannot pass at this moment, it is fine to merge since at least it has the set up.

Added! It should fail though.

tianleh (approved):

qreshi:
It looks like all the other test workflows are checking out version 1.3.0 of OpenSearch. Are those going to be bumped to 2.0 as well?

qreshi:
Also, is this PR copying all of the Cypress tests in the Notifications repo into this one? Can we not checkout the Notifications repo and run the Cypress tests that way? This is duplicating the code and if we change the tests on our side it wouldn't reflect here.

By the way, the Cypress tests for Notifications are failing right now (as in they need to be updated due to changes in the frontend in preparation for release). For the time being, only the Jest unit tests are stable.

kavilla:
> It looks like all the other test workflows are checking out version 1.3.0 of OpenSearch. Are those going to be bumped to 2.0-rc1 as well?

Probably should but I will defer to @tianleh for that notifications plugin didn't exist before so couldn't use 1.3.0, am I correct?

> Also, is this PR copying all of the Cypress tests in the Notifications repo into this one? Can we not checkout the Notifications repo and run the Cypress tests that way? This is duplicating the code and if we change the tests on our side it wouldn't reflect here.

You can the ideal here is that just to get notifications plugin onboard and tested against a full distribution with other tests, a sanity tests that can differ from the integration tests within the plugin repo. So a full E2E test. Albeit there is redundancy feel free to modify the tests post onboard so we know how this plugin impacts the rest of of the distribution. Another added benefit of onboarding here is that if there is any library breaking changes that come upstream you wouldn't have to worry about it as much as maintaining your own distribution.

That said if you only rely on unit tests how confident are you in the actual plugin? How will you verify it?

qreshi:
No, what I mean is, we keep the Cypress tests in our own repo and I'm pretty sure every other repo does too. They are usually run locally and as part of the workflow for the repo in the PR.

If we are copying those tests here, then it's a duplication since we have them in our repo and they could change over time (we'll lose the benefit of the new tests added there being run here). If we remove them from the repo, then that puts our tests in two different places which I'm not a fan of.

I think the long term solution should be to have the repo's checked out when the Cypress tests are run for each installed plugin (looks like [this](https://github.com/opensearch-project/opensearch-dashboards-functional-test/blob/main/.github/workflows/release-e2e-workflow-template.yml) is the workflow to change for that?) so it's still run against the full distribution but there is a single source of truth for the test definition. I'm curious why we didn't do it that way from the beginning? Are there any issues with that approach that I'm not aware of?

> That said if you only rely on unit tests how confident are you in the actual plugin in a distribution? How will you verify it? Does the security plugin impact it?

For Notifications, the Cypress test broke as we were making changes and haven't had a chance to update them yet since we're trying to get ready for release. We don't plan on only relying on Jest tests, I just wanted to let you know that if we run them for Notifications as they currently are, it might fail.

kavilla:
> If we are copying those tests here, then it's a duplication since we have them in our repo and they could change over time (we'll lose the benefit of the new tests added there being run here). If we remove them from the repo, then that puts our tests in two different places which I'm not a fan of.

I understand, I think just to be clear the point of copying the tests here is just so there is some tests running for the full distribution tests. These don't have to be the final draft of the tests but just something. If you'd prefer to manage your own you will need to do this for every release: https://github.com/opensearch-project/opensearch-build/blob/main/ONBOARDING.md#onboard-to-test-workflow.

I think the confusion is that we should consider this repo as a sanity test repo for a full E2E experience. Where as having each individual plugin ran individually in it's own controlled environment. 

Another issue is that once OpenSearch Dashboards onboards cypress as it's official test framework OpenSearch Dashboards plugins will now have conflicts since not all plugins are on the same version but will conflict with the dependency tree that it will inherit from OSD.

All this coupled with us handling any issue with an node upgrade incompatibility, it would be less work for your repo if onboard some sanity tests into this repo and have integrations test running per PR in your repo. But if you prefer then you will need to make sure for `2.0.0-rc1` you onboard or run manually.

seraphjiang (approved):

seraphjiang:
Let's merge and ask for plugin team to fix flake test

CC: @xinlamzn  @praveensameneni

peterzhuamazon:
What's the plan to merge since we need automation.
https://github.com/opensearch-project/notifications/issues/433

Can we merge this PR and ask plugin owners to continue fix this?

Thanks.

tianleh:
> What's the plan to merge since we need automation. [opensearch-project/notifications#433](https://github.com/opensearch-project/notifications/issues/433)
> 
> Can we merge this PR and ask plugin owners to continue fix this?
> 
> Thanks.

Merging. As @qreshi mentioned above, there might be some flakeness about the tests.

PR #227, opened 2022-05-16
AD test failure fix 

amitgalitz:
Signed-off-by: Amit Galitzky <amgalitz@amazon.com>

### Description

Added wait time to solve AD test looking for something in the next page after button click.
### Check List

### DIFF
--- a/cypress/integration/plugins/anomaly-detection-dashboards-plugin/create_detector_spec.js
+++ b/cypress/integration/plugins/anomaly-detection-dashboards-plugin/create_detector_spec.js
@@ -50,6 +50,8 @@ context('Create detector workflow', () => {
     cy.getElementByTestId('indicesFilter').type(`${TEST_INDEX_NAME}{enter}`);
     selectTopItemFromFilter('timestampFilter', false);
     cy.getElementByTestId('defineDetectorNextButton').click();
+    // wait for a few seconds to move to the next page
+    cy.wait(3000);
     cy.getElementByTestId('defineOrEditDetectorTitle').should('not.exist');
     cy.getElementByTestId('configureOrEditModelConfigurationTitle').should(
       'exist'
--- a/cypress/integration/plugins/anomaly-detection-dashboards-plugin/create_detector_spec.js
+++ b/cypress/integration/plugins/anomaly-detection-dashboards-plugin/create_detector_spec.js
@@ -49,11 +49,8 @@ context('Create detector workflow', () => {
     );
     cy.getElementByTestId('indicesFilter').type(`${TEST_INDEX_NAME}{enter}`);
     selectTopItemFromFilter('timestampFilter', false);
-    cy.wait(500);
 
     cy.getElementByTestId('defineDetectorNextButton').click();
-    // wait for a few seconds to move to the next page
-    cy.wait(3000);
     cy.getElementByTestId('defineOrEditDetectorTitle').should('not.exist');
     cy.getElementByTestId('configureOrEditModelConfigurationTitle').should(
       'exist'
--- a/cypress/integration/plugins/anomaly-detection-dashboards-plugin/create_detector_spec.js
+++ b/cypress/integration/plugins/anomaly-detection-dashboards-plugin/create_detector_spec.js
@@ -49,6 +49,8 @@ context('Create detector workflow', () => {
     );
     cy.getElementByTestId('indicesFilter').type(`${TEST_INDEX_NAME}{enter}`);
     selectTopItemFromFilter('timestampFilter', false);
+    cy.wait(500);
+
     cy.getElementByTestId('defineDetectorNextButton').click();
     // wait for a few seconds to move to the next page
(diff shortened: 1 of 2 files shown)


cypress/integration/plugins/anomaly-detection-dashboards-plugin/create_detector_spec.js
```diff
@@ -49,7 +49,11 @@ context('Create detector workflow', () => {
     );
     cy.getElementByTestId('indicesFilter').type(`${TEST_INDEX_NAME}{enter}`);
     selectTopItemFromFilter('timestampFilter', false);
+    cy.wait(500);
```
kavilla, line 52: 😭 

Is there an API call that you can intercept and wait until that finishes to continue?

If not, we can also extend the functionality of https://github.com/opensearch-project/opensearch-dashboards-functional-test/blob/25afb705b4d24413a330cde51f1e74803555ecc7/cypress/utils/commands.js#L70 to allow for `{ timeout: xxxx }` for cy.get [[doc](https://docs.cypress.io/api/commands/get#Timeouts)]. That way it will navigate to the page and check for longer that the default amount. But if it finds it earlier then it will continue.

amitgalitz: The second option makes sense to me, I don't think there is an API call that we can intercept easily. However I am still trying to figure out why this really sometimes helps but its pretty hard to replicate locally.

kavilla: > However I am still trying to figure out why this really sometimes helps but its pretty hard to replicate locally.

With the second option? Or just adding a timeout in general?

kavilla: Sorry, just to clarify I mean extending it in the way that makes the timeout configurable. RIght now it has the default timeout and when you pass that command if you increase for example to 1 minute (`{ timeout: 60000}`). It will tell cypress to keep retrying for 1 minute to find the element. But if it finds it in 3 seconds it will continue the test path and not wait for the whole minute. The point where it will increase the time for all tests if all tests actually end up failing to find the element but generally all tests shouldn't experience an increase wait time. This is compared to the arbitrary wait for 3 seconds where it will wait for 3 seconds then move to the next line. But if it actually required 4 seconds then the test will still fail.

amitgalitz: I moved the wait time inside this [function](https://github.com/opensearch-project/opensearch-dashboards-functional-test/blob/main/cypress/utils/plugins/anomaly-detection-dashboards-plugin/helpers.js#L6) as it seemed to help with the issue the most, since it isn't part of `getElementByTestId` do you recommend moving the wait to the next line of: https://github.com/opensearch-project/opensearch-dashboards-functional-test/pull/227/files#diff-1544ad6eba828cbcff2b11e02c24a29dfa194e5da1cc11de42463f46c3511073R53. For adding the functionality since default is 60s, it would have to be passed with 61000? It doesn't seem to be dependent on the 60 second wait time itself since just 1 second more or even .5 seconds solves the issue which makes me wary of just increasing that wait time. It more seems to be the transition between this line (`cy.get('.euiFilterSelectItem').first().click();`) and then clicking the next.

amitgalitz:
Test consistently passes on github runner now on latest commit

cypress/integration/plugins/anomaly-detection-dashboards-plugin/create_detector_spec.js
ohltyler: Agree with @amitgalitz here, the issue is related to not enough time within `selectTopItemFromFilter` to load the data and perform the action, such that it gets hung and the runner is not able to click to the next page regardless of the 60s timeout. I think the fix makes sense here

ohltyler (approved):

jackiehanyang (approved):

cypress/integration/plugins/anomaly-detection-dashboards-plugin/create_detector_spec.js
tianleh: Per Cypress [guidance](https://docs.cypress.io/api/commands/wait#Time), it is anti-pattern to have an explicit wait. Totally understand that this has already been used in other test cases. Just try to see if this is a low hanging fruit to fix. By checking more into the Cypress best [practice](https://docs.cypress.io/guides/references/best-practices#Unnecessary-wait-for-cy-get), I think we can add an assertion for the `get` which can function as waiting and then try to chain into it.

@amitgalitz @ohltyler

amitgalitz: Thats a fair point, I'll send an additional PR this week removing the explicit wait and hopefully other uses of it in AD tests

PR #230, opened 2022-05-19
Wait for tenancy call if security enabled and bump version

kavilla:
### Description

Bumping to version 2.0.0, fix linter problem.

Also, preventing the test runner from navigating to a new
page when security plugin is enabled. The reason why is
because depending on how long the test takes to run, there
is an API call to refresh the permissions that take a little
bit longer than what tests are expecting. So if the test runner
makes the call to switch tenants but navigates to a new page
before that call is finished another call might run into a
permissions error. This ensures that if that call is made
then the test runner waits before navigating to the destination.

### Issues Resolved

https://github.com/opensearch-project/opensearch-dashboards-functional-test/issues/231

### Check List

### DIFF
--- a/.github/workflows/cypress-workflow-bundle-snapshot-based.yml
+++ b/.github/workflows/cypress-workflow-bundle-snapshot-based.yml
@@ -55,7 +55,7 @@ jobs:
         run: |
           cd opensearch-dashboards-${{ env.VERSION }}
           bin/opensearch-dashboards serve &
-          timeout 300 bash -c 'while [[ "$(curl -s localhost:5601/api/status | jq -r '.status.overall.state')" != "green" ]]; do sleep 5; done'
+          timeout 300 bash -c 'while [[ "$(curl -s -u admin:admin -k localhost:5601/api/status | jq -r '.status.overall.state')" != "green" ]]; do sleep 5; done'
       - name: Get Cypress version
         id: cypress_version
         run: |
--- a/.github/workflows/release-e2e-workflow-template.yml
+++ b/.github/workflows/release-e2e-workflow-template.yml
@@ -55,7 +55,7 @@ jobs:
         run: |
           cd opensearch-dashboards-${{ env.VERSION }}
           bin/opensearch-dashboards serve &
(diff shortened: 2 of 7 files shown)


kavilla:
Failures would be due to chicken and egg problem. Needs a 2.0 build but this needs to be in the 2.0 build.

tianleh:
> Failures would be due to chicken and egg problem. Needs a 2.0 build but this needs to be in the 2.0 build.

By checking the error in this check, https://github.com/opensearch-project/opensearch-dashboards-functional-test/runs/6514651442?check_suite_focus=true ,  the error is 

```
{"type":"response","@timestamp":"2022-05-19T20:43:07Z","tags":["api"],"pid":2089,"method":"get","statusCode":401,"req":{"url":"/api/status","method":"get","headers":{"host":"localhost:5601","user-agent":"curl/7.68.0","accept":"*/*"},"remoteAddress":"127.0.0.1","userAgent":"curl/7.68.0"},"res":{"statusCode":401,"responseTime":1,"contentLength":9},"message":"GET /api/status 401 1ms - 9.0B"}
```

It is related to the breaking change where `/api/status` needs authentication.( https://github.com/opensearch-project/opensearch-build/issues/2112#issuecomment-1125254886)

@kavilla Since you already have the version bump in this PR, can you add such authentication along the way? We already have an existing [example](https://github.com/opensearch-project/opensearch-dashboards-functional-test/blob/main/.github/workflows/release-e2e-workflow-template.yml#L40) for 9200 check .

cc @seraphjiang @CCongWang @peterzhuamazon

tianleh (approved):

kavilla:
@tianleh seeing some issues. but might be related to the current existing failures from AD and Observability. The other PR might help but I believe both plugins are working on it.

peterzhuamazon:
@kavilla @tianleh I vote for merging this PR now:
1. OBS #216 is waiting for this PR to merge.
2. ADDash is also waiting for this to verify their test again.
Thanks.

tianleh:
> @kavilla @tianleh I vote for merging this PR now:
> 
>     1. OBS

+1

PR #295, opened 2022-09-28
Change form:false for AD request

ohltyler:

### Description

Changes the AD request to `form:false` in the request parameters, when loading in some sample data. From the [cypress documentation](https://docs.cypress.io/api/commands/request#Arguments), setting to `form:true` means the body should be url-encoded, and overwrites the `Content-Type` header to `application/x-www-form-urlencoded`. What we actually want is `application/json` which we have already specified in the request itself. By setting to `false` it removes any confusion and cleans up the code to make it run as expected.

Note that the previous code still worked, but was changing the request header to `Content-Type: application/x-www-form-urlencoded` under the hood which was causing confusion and could lead to certain bugs, especially if the request body was changed such that it couldn't be url-encoded.

### Check List

### DIFF
--- a/cypress/integration/plugins/anomaly-detection-dashboards-plugin/create_detector_spec.js
+++ b/cypress/integration/plugins/anomaly-detection-dashboards-plugin/create_detector_spec.js
@@ -19,7 +19,7 @@ context('Create detector workflow', () => {
     cy.fixture(AD_FIXTURE_BASE_PATH + 'sample_test_data.txt').then((data) => {
       cy.request({
         method: 'POST',
-        form: true,
+        form: false,
         url: 'api/console/proxy',
         headers: {
           'content-type': 'application/json;charset=UTF-8',


amitgalitz (approved):

tianleh:
Looks like that there are more places with such https://github.com/opensearch-project/opensearch-dashboards-functional-test/blob/main/cypress/integration/plugins/gantt-chart-dashboards/gantt_ui.spec.js#L22

I am ok if we just focus on one plugin in this PR. @CCongWang

tianleh:
side note: the bundle [workflow](https://github.com/opensearch-project/opensearch-dashboards-functional-test/actions/runs/3146020229/jobs/5113990645) failed because there is no real OS plugin yet.

https://github.com/opensearch-project/opensearch-build/blob/main/manifests/3.0.0/opensearch-3.0.0.yml

tianleh (approved):

CCongWang (approved):

tianleh:
Some branches failed to backport. @ohltyler  @CCongWang

PR #324, opened 2022-10-12
[CI] Update test template to not use timeout

kavilla:
### Description

GitHub workflow was failing because starting of OpenSearch and therefore OpenSearch Dashboards would exit with 124.

For example:
https://github.com/opensearch-project/opensearch-dashboards-functional-test/actions/runs/3237324644/jobs/5304250104

This is due to us using a timeout function and it outputs a non-zero code therefore github actions treats it as a failure code and doesn't run the test.

This fixes to utilize cypress to run the startup commands and tests within a single command.

### Issues Resolved

n/a

### Check List

### DIFF
--- a/.github/workflows/release-e2e-workflow-template.yml
+++ b/.github/workflows/release-e2e-workflow-template.yml
@@ -75,7 +75,7 @@ jobs:
           working-directory: cypress-test
           command: ${{ inputs.test-command }}
           start: ${{ steps.get_opensearch.RUN_CMD }}, ${{ steps.get_opensearch_dashboards.RUN_CMD }}
-          wait-on: 'http://localhost:9200, http://localhost:5601'
+          wait-on: 'https://localhost:9200, http://localhost:5601'
       # Screenshots are only captured on failure, will change this once we do visual regression tests
       - uses: actions/upload-artifact@v1
         if: failure()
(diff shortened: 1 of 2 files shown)


.github/workflows/release-e2e-workflow-template.yml
```diff
@@ -78,7 +74,8 @@ jobs:
         with:
           working-directory: cypress-test
           command: ${{ inputs.test-command }}
-          wait-on: 'http://localhost:5601'
+          start: ${{ steps.get_opensearch.RUN_CMD }}, ${{ steps.get_opensearch_dashboards.RUN_CMD }}
+          wait-on: 'http://localhost:9200, http://localhost:5601'
```
tianleh, line 78: In bundled OSD (with security enabled), the 9200 is under `https` not `http`.

kavilla: shoot good catch

tianleh:
@kavilla Do we still need this PR?

kavilla:
> @kavilla Do we still need this PR?

What is your opinion? Does it make it cleaner? It appears that the original issue was due to something else but I like the non-ambiguous timeout step.

If so I can add the condition for passing the security params.

tianleh:
No longer needed. We will revisit in the future.

PR #331, opened 2022-10-17
Fix flaky clicks on AD historical tests

ohltyler:

### Description

Adds wait times before some of the click actions on some of the buttons for the historical detection tests in the anomaly detection plugin test suite.

We have seen flaky failures where some of the buttons are not clicked when expected, leading to confusing errors later on due to the page getting into an unexpected state. This is because the actions taken right before the click are making API calls and reloading part of the page, so the click doesn't happen. To fix this, we add wait times before trying to click. Also adding `force` params to some of the click actions for consistency across the tests and to prevent future potentially flaky failures.

### Check List

### DIFF
--- a/cypress/integration/plugins/anomaly-detection-dashboards-plugin/historical_analysis_spec.js
+++ b/cypress/integration/plugins/anomaly-detection-dashboards-plugin/historical_analysis_spec.js
@@ -135,6 +135,7 @@ describe('Historical results page', () => {
         force: true,
       });
       cy.get(`[aria-label="Next time window"]`).click();
+      cy.wait(5000);
       cy.contains('Refresh').click({ force: true });
       verifyNoAnomaliesInCharts();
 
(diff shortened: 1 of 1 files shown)


amitgalitz:
Do you know if there is maybe an api we can intercept and wait for it to finish specifically or why waiting for a specific element ID first doesn't work?

ohltyler:
> Do you know if there is maybe an api we can intercept and wait for it to finish specifically or why waiting for a specific element ID first doesn't work?

The element ID is always present - the issue is that when the previous cmd makes an async call, but the `click` action to refresh happens immediately, it's undeterministic on if the `click` will happen during or after the async call is finished, which is when the component that contains the 'Refresh' button is re-rendered. By just waiting 5s for the calls to finish and the page to be stable, we can deterministically click on it. It's technically not just about waiting for the call to finish, but rather the call to finish _and_ page to re-render, which doesn't seem to work well without adding some manual wait times.

tianleh:
Shall this backport to `2.x`?

tianleh (approved):

PR #547, opened 2023-02-21
Fix Sql timeout

kavithacm:
### Description

Fix Sql timeout

### Issues Resolved

### Check List

### DIFF
--- a/cypress/integration/plugins/query-workbench-dashboards/ui.spec.js
+++ b/cypress/integration/plugins/query-workbench-dashboards/ui.spec.js
@@ -99,7 +99,7 @@ describe('Test PPL UI', () => {
       .type('source=accounts', { force: true });
     cy.wait(QUERY_WORKBENCH_DELAY);
     cy.get('.euiButton__text').contains('Run').click({ force: true });
-    cy.wait(QUERY_WORKBENCH_DELAY);
+    cy.wait(QUERY_WORKBENCH_DELAY * 5);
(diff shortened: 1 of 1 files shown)


CCongWang (approved):

cypress/integration/plugins/query-workbench-dashboards/ui.spec.js
```diff
@@ -182,8 +182,8 @@ describe('Test SQL UI', () => {
     cy.get('.euiTitle').contains('Query Workbench').should('exist');
 
     cy.get('.euiButton__text').contains('Run').click({ force: true });
-    cy.wait(QUERY_WORKBENCH_DELAY);
-    cy.get('.euiButton__text', { withinSubject: null })
```
joshuali925, line 186: why remove this?

CCongWang: yeah I just tested it, so I was wrong before.. I think the main issue is that, after it clicks Run, it just waits for 1s, and `Full screen view` fails to show up after 1s.
In my test, I removed `{ withinSubject: null }` and add more wait time, tests passed

joshuali925 (approved):

cypress/integration/plugins/query-workbench-dashboards/ui.spec.js
joshuali925: i see, thanks for checking

PR #552, opened 2023-02-21
Add Cypress test for Default Saved Map

naveentatikonda:
### Description

Add Cypress test to validate adding default saved map for the sample flights dataset in the dashboards-maps plugin.

### Check List

### DIFF
--- a/cypress/integration/plugins/custom-import-map-dashboards/add_saved_object.spec.js
+++ b/cypress/integration/plugins/custom-import-map-dashboards/add_saved_object.spec.js
@@ -1,0 +1,40 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { BASE_PATH } from '../../../utils/constants';
+
+describe('Add flights dataset saved object', () => {
+  before(() => {
+    cy.visit(`${BASE_PATH}/app/maps-dashboards`, {
+      retryOnStatusCodeFailure: true,
+      timeout: 60000,
+    });
+    cy.get('div[data-test-subj="indexPatternEmptyState"]', { timeout: 60000 })
+      .contains(/Add sample data/)
+      .click();
+    cy.get('div[data-test-subj="sampleDataSetCardflights"]', { timeout: 60000 })
+      .contains(/Add data/)
+      .click();
+    cy.wait(60000);
+  });
+
+  after(() => {
+    cy.visit(`${BASE_PATH}/app/home#/tutorial_directory`);
+    cy.get('button[data-test-subj="removeSampleDataSetflights"]')
+      .should('be.visible')
+      .click();
+  });
+
+  it('check if maps saved object of flights dataset can be found and open', () => {
+    cy.visit(`${BASE_PATH}/app/maps-dashboards`);
+    cy.contains(
+      '[Flights] Flights Status on Maps Destination Location'
+    ).click();
+    cy.get('[data-test-subj="layerControlPanel"]').should(
+      'contain',
+      'Flights On Time'
+    );
+  });
+});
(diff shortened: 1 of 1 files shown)


kavilla:
Can you do me a favor in **seperate** PR can you add your test to the [test results explorer site](https://opensearch-project.github.io/opensearch-dashboards-functional-test/site/) for easier debugging in the future:
* https://github.com/opensearch-project/opensearch-dashboards-functional-test/blob/main/site/js/dashboard.js#L16
* https://github.com/opensearch-project/opensearch-dashboards-functional-test/blob/main/site/index.html#L134

cypress/integration/plugins/custom-import-map-dashboards/add_saved_object.spec.js
```diff
@@ -0,0 +1,40 @@
...
+      retryOnStatusCodeFailure: true,
+      timeout: 60000,
+    });
+    cy.get('div[data-test-subj="indexPatternEmptyState"]', { timeout: 60000 })
+      .contains(/Add sample data/)
+      .click();
+    cy.get('div[data-test-subj="sampleDataSetCardflights"]', { timeout: 60000 })
+      .contains(/Add data/)
+      .click();
+    cy.wait(60000);
```
kavilla, line 20: This should be avoided. This is an arbitrary wait time that guarantees we add a minute to the test run. I would utilize the timeout feature to wait for something that loads on the page that way we are waiting until an element loads instead of guessing how long it takes.

kavilla (commented):
My suggestion as well is add a number before the file to ensure the order.

cypress/integration/plugins/custom-import-map-dashboards/add_saved_object.spec.js
naveentatikonda: ack

naveentatikonda:
> My suggestion as well is add a number before the file to ensure the order.

I believe for dashboards-maps tests order doesn't matter because they are all running independently.

kavilla:
> > My suggestion as well is add a number before the file to ensure the order.
> 
> I believe for dashboards-maps tests order doesn't matter because they are all running independently. But, if you want to follow that naming convention then I will create a separate PR renaming the tests in dashboards-maps

With hindsight, I've seen developers get test failures. For scalability when more developers contribute they might assume that a specific test will always come before another test and write tests in such a fashion (which isn't great but not preventable) but then they add a new file and since cypress will run in alpha-numeric order and failures occur. So adding a number ensures tests run in a specific order.

naveentatikonda:
> > > My suggestion as well is add a number before the file to ensure the order.
> > 
> > 
> > I believe for dashboards-maps tests order doesn't matter because they are all running independently. But, if you want to follow that naming convention then I will create a separate PR renaming the tests in dashboards-maps
> 
> With hindsight, I've seen developers get test failures. For scalability when more developers contribute they might assume that a specific test will always come before another test and write tests in such a fashion (which isn't great but not preventable) but then they add a new file and since cypress will run in alpha-numeric order and failures occur. So adding a number ensures tests run in a specific order.

Got it. Will raise a separate PR renaming all dashboards-maps tests by adding numbers

kavilla:
forgot to mention you need to add your test here: https://github.com/opensearch-project/opensearch-dashboards-functional-test/blob/main/test_finder.sh#L10

junqiu-lei:
> forgot to mention you need to add your path to your tests and your plugin id: https://github.com/opensearch-project/opensearch-dashboards-functional-test/blob/main/test_finder.sh#L10

Hi @kavilla, the layered maps App is still under custom-import-map-dashboards plugin. I don't think we need change here. https://github.com/opensearch-project/opensearch-dashboards-functional-test/blob/main/test_finder.sh#L20

kavilla:
> > forgot to mention you need to add your path to your tests and your plugin id: https://github.com/opensearch-project/opensearch-dashboards-functional-test/blob/main/test_finder.sh#L10
> 
> Hi @kavilla, the layered maps App is still under custom-import-map-dashboards plugin. I don't think we need change here. https://github.com/opensearch-project/opensearch-dashboards-functional-test/blob/main/test_finder.sh#L20

whoops missed it! thank you sorry about that

kavilla:
is the goal is to get this to 2.x branch?

naveentatikonda:
> is the goal is to get this to 2.x branch?

Yes, actually to 2.6

kavilla (approved):

CCongWang (approved):

PR #380, opened 2022-11-07
Onboarding security analytics dashboards plugin.

AWSHurneyt:
### Description
Onboarding security analytics dashboards plugin. Security analytics 3.0 isn't available yet, so submitting this PR directly to the function test repo 2.x branch.

### Issues Resolved

[List any issues this PR will resolve]

### Check List

- [x] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #591, opened 2023-03-17
Onboarding security analytics dashboards plugin. (#380)

kavilla:
Forwardport of 165f8559cca46f141fe9460463b6793ca4b14ad5 from https://github.com/opensearch-project/opensearch-dashboards-functional-test/pull/380


kavilla:
Even though the 3.0 plugin version isn't available yet. It's making branching strategy difficult that it's not work was only done on 2.5

kavilla:
I think I should lint checker fail because I need to forwardport another commit that handles that file and the conflicts might be plenty.

kavilla:
Closing as duplicate.

PR #644, opened 2023-04-26
Add an optional parameter -o to integtest.sh 

Divyaasm:
SignedOff by : divyaasm@amazon.com

### Description

A new optional argument OPTION [-o ] is added to integtest.sh

### Issues Resolved

Fix to the issue [#3316](https://github.com/opensearch-project/opensearch-build/pull/3316)

Call to ./integtest.sh from integ-workflow can now use additional `-o default` parameter when -t is passed which executes the tests individually instead of altogether. This is optional and does not change the behaviour if -o is not passed.

### Check List

### DIFF
--- a/integtest.sh
+++ b/integtest.sh
@@ -2,8 +2,6 @@
 
 set -e
 
-. ./test_finder.sh
-
 function usage() {
     echo ""
     echo "This script is used to run integration tests for plugin installed on a remote OpenSearch/Dashboards cluster."
@@ -20,11 +18,12 @@ function usage() {
     echo -e "-c CREDENTIAL\t(usename:password), no defaults, effective when SECURITY_ENABLED=true."
     echo -e "-t TEST_COMPONENTS\t(OpenSearch-Dashboards reportsDashboards etc.), optional, specify test components, separate with space, else test everything."
     echo -e "-v VERSION\t, no defaults, indicates the OpenSearch version to test."
+    echo -e "-o OPTION\t, no defaults, determine the TEST_TYPE value among(default, manifest) in test_finder.sh, optional."
     echo -e "-h\tPrint this message."
     echo "--------------------------------------------------------------------------"
 }
 
-while getopts ":hb:p:s:c:t:v:" arg; do
+while getopts ":hb:p:s:c:t:v:o:" arg; do
     case $arg in
         h)
             usage
@@ -47,6 +46,9 @@ while getopts ":hb:p:s:c:t:v:" arg; do
             ;;
         v)
             VERSION=$OPTARG
+            ;;
+        o)
+            OPTION=$OPTARG
             ;;
         :)
             echo "-${OPTARG} requires an argument"
@@ -82,6 +84,8 @@ then
   USERNAME=`echo $CREDENTIAL | awk -F ':' '{print $1}'`
   PASSWORD=`echo $CREDENTIAL | awk -F ':' '{print $2}'`
 fi
+
+. ./test_finder.sh
 
 npm install
 
--- a/test_finder.sh
+++ b/test_finder.sh
@@ -5,6 +5,7 @@ set -e
 OSD_BUILD_MANIFEST='../local-test-cluster/opensearch-dashboards-*/manifest.yml'
 OSD_TEST_PATH='cypress/integration/core-opensearch-dashboards'
 OSD_PLUGIN_TEST_PATH='cypress/integration/plugins'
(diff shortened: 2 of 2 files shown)


integtest.sh
```diff
@@ -20,11 +18,12 @@ function usage() {
     echo -e "-c CREDENTIAL\t(usename:password), no defaults, effective when SECURITY_ENABLED=true."
     echo -e "-t TEST_COMPONENTS\t(OpenSearch-Dashboards reportsDashboards etc.), optional, specify test components, separate with space, else test everything."
     echo -e "-v VERSION\t, no defaults, indicates the OpenSearch version to test."
+    echo -e "-o OPTION\t, no defaults, determine the TEST_TYPE value among(default, manifest) in test_finder.sh, optional."
```
ashwin-pc, line 21: why dont we want a default value here? since one of the accepted value is already `default`

rishabh6788: The background of the issue is that when you the run the OSD integ test by simply checking out the FT repo and directly invoking `integtest.sh` script the TEST_TYPE is `default` and if you look at `test_finder.sh` script, in case the TEST_TYPE is `default` it acknowledges the component passed using `-t` flag and runs the test for that particular component. 

But in case the OSD integ test is triggered using the integ-test [workflow](https://github.com/opensearch-project/opensearch-build/blob/main/test.sh) the TEST_TYPE is `manifest` and the `test_finder.sh` script doesn't acknowledge the passed component and runs test for all the components. 

At this moment we don't know the logic behind the TEST_TYPE field and this PR is to force set the TEST_TYPE to `default` when running from integ-test workflow so that when we pass the specific component the integ tests are run for that only. 
This change doesn't alter the current behavior is in case the TEST_TYPE is not force set it will follow the current behavior. 
Hope it makes sense.  
@ashwin-pc

rishabh6788 (approved):
LGTM!

integtest.sh
gaiksaya: I think its the wording that @ashwin-pc is saying. If you do not pass `-o` test_finder.sh will anyway have `TEST_TYPE = default`

ashwin-pc: @rishabh6788 thanks for the context. Yeah as @gaiksaya points out, my comment was about the wording `no defaults` but based on your comment the `-o` flag is an optional flag that can take a `TEST_TYPE` value, but when not present defaults to `default`. The echo statement here does not say that and makes it seem like the `-o` flag is a required flag with no defaults. As a caveat, im not sure what the `TEST_TYPE` flag does either and my comment is strictly about the instructions provided to a user of the script when we echo the help instructions.

Divyaasm: As per my testing I see TEST_TYPE had null value when we directly call the integtest script inside a cloned repo, which is later getting initialized at [line](https://github.com/opensearch-project/opensearch-dashboards-functional-test/blob/9e838bb9ba4b2c64ed2fb8337cde9e522824bab2/test_finder.sh#L26). But not sure if were are setting the variable TEST_TYPE when the script is called from a different source.

That's the reason we didn't alter it's value if TEST_TYPE is already set.

gaiksaya: Hi @ashwin-pc, @Divyaasm  will be creating the issue for fixing the bug here:https://github.com/opensearch-project/opensearch-dashboards-functional-test/blob/main/test_finder.sh#L57-L75
If manifest is provided there is no option to run integration test for just 1 component today. So we will be using `default` now. If you think `TEST_TYPE` manifest should be removed completely that would work as well. I believe that was added for same distribution integration test which is not being used now. 

Thanks!

kavilla (approved):
Seems to fine me.

Historically there was no ability to run this repo and specific tests from the Jenkins CI for OSD. So this script was to added to prevent the default run where it scanned the directory of integration tests and just ran them based on if it's either default (so all the existing tests) or checking if the plugin existed.

So I think it's good but it is still open to user error where if they pass in the test type to be default and then pass a test component that doesn't exist in the build then the tests will fail. But that can be an improvement later if it's enough of a problem.

gaiksaya:
Here is the bug issue created @ashwin-pc @kavilla https://github.com/opensearch-project/opensearch-dashboards-functional-test/issues/648

Thanks!

prudhvigodithi:
@Divyaasm can you please backport this to `1.3` branch to move forward with `1.3.10` release.
https://github.com/opensearch-project/opensearch-build/issues/3331

PR #676, opened 2023-05-25
Update vanilla snapshot test to 2.7.0 and replace deprecated set-output

peterzhuamazon:
### Description

Update vanilla snapshot test to 2.7.0 and replace deprecated set-output

### Issues Resolved

https://github.com/opensearch-project/opensearch-build/pull/3434

### Check List


peterzhuamazon:
Also removed two disabled workflows that have not being used for a long time.

CCongWang (approved):

peterzhuamazon:
Need to work with @RyanL1997 to fix the data table then have the version dynamically change based on `version` in package.json.

RyanL1997:
@peterzhuamazon (cc @ananzh )
### Here are some of the findings about the failure of `data tables` test case:
**OS version 2.7 with Cypress 9**
issuesOn2 7Cy9

**OS version 2.0 with Cypress 9**
issuesOn2 0Cy9

It seems like we are missing this `paginated-table-body` for the latest OS version.

peterzhuamazon:
> @peterzhuamazon (cc @ananzh )
> 
> **Here are some of the findings about the failure of `data tables` test case:** **OS version 2.7 with Cypress 9** issuesOn2 7Cy9
> 
> **OS version 2.0 with Cypress 9** issuesOn2 0Cy9
> 
> It seems like we are missing this `paginated-table-body` for the latest OS version.

Thanks @RyanL1997 for looking this up.
Meanwhile, I have pushed a change to auto grab the version defined in package.json.
So that this test can properly install the corresponding versions instead of hardcoding to 2.0.0 or others.

Thanks.

kavilla:
Table fix incoming

PR #787, opened 2023-07-22
[Table Visualizations] Test Update

curq:
### Description

In this dashboards [PR](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/4272) we replaced div container with `OuiFlexItem`. So this line:
https://github.com/opensearch-project/opensearch-dashboards-functional-test/blob/7f8a7d799abe560549759fa440b0184a3238b61b/cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/vis_builder/vis_types/table.spec.js#L91
Was causing error, because it matches element with exactly single `visTable__group` class, while after the PR we had additional `euiFlexItem` class. As a result we got undefined. Here is the screenshot of error.
Vis Builder Table Chart -- Basic test (failed)

Now after this fix this test suite do pass, here is the screenshot from my local cypress run.
image

As we can see it does match all 4 elements like it should.

### Issues Resolved

It solves cypress error when trying to merge https://github.com/opensearch-project/OpenSearch-Dashboards/pull/4272

### Check List

### DIFF
--- a/cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/vis_builder/vis_types/table.spec.js
+++ b/cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/vis_builder/vis_types/table.spec.js
@@ -88,7 +88,7 @@ export const removeBucket = (bucket) => {
 export const testSplitTables = (num) => {
   cy.getElementByTestId('visTable')
     .should('have.class', 'visTable')
-    .find('[class="visTable__group"]')
+    .find('[class*="visTable__group"]')
     .should(($tables) => {
       // should have found specified number of tables
       expect($tables).to.have.length(num);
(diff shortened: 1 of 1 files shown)


cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/vis_builder/vis_types/table.spec.js
```diff
@@ -88,7 +88,7 @@ export const removeBucket = (bucket) => {
 export const testSplitTables = (num) => {
   cy.getElementByTestId('visTable')
     .should('have.class', 'visTable')
-    .find('[class="visTable__group"]')
+    .find('[class*="visTable__group"]')
```
ruanyl, line 91: Nit: can it simply be `find('.visTable__group')`?

ruanyl (approved):

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/vis_builder/vis_types/table.spec.js
curq: Yes, I tested it now and it works
image
I have updated PR to simpler version you suggested

PR #865, opened 2023-09-15
[Alerting 2.10] Override element covered check for name input in Composite monitor creation

amsiglan:
### Description
We are seeing flaky behavior when creating composite monitor. The name input element is sometimes not in view which prevents cypress from typing the name. Since this behavior won't happen when a user is creating the monitor it is safe to force the typing in the name element and the monitor selector elements.

### Check List

### DIFF
--- a/cypress/integration/plugins/alerting-dashboards-plugin/composite_level_monitor_spec.js
+++ b/cypress/integration/plugins/alerting-dashboards-plugin/composite_level_monitor_spec.js
@@ -71,16 +71,20 @@ describe('CompositeLevelMonitor', () => {
       cy.get('[data-test-subj="visualEditorRadioCard"]').click({ force: true });
 
       // Wait for input to load and then type in the monitor name
-      cy.get('input[name="name"]').type(SAMPLE_VISUAL_EDITOR_MONITOR);
+      cy.get('input[name="name"]').type(SAMPLE_VISUAL_EDITOR_MONITOR, {
+        force: true,
+      });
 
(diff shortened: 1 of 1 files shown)


cypress/integration/plugins/alerting-dashboards-plugin/composite_level_monitor_spec.js
```diff
@@ -71,26 +71,30 @@ describe('CompositeLevelMonitor', () => {
...
       cy.get('[title="monitorTwo"]').click({ force: true });
 
       cy.get('button').contains('Add trigger').click({ force: true });
 
       // Type trigger name
       cy.get('[data-test-subj="composite-trigger-name"]')
-        .type('{selectall}')
-        .type('{backspace}')
-        .type('Composite trigger');
+        .type('{selectall}', { force: true })
```
kavilla, line 95: in the past I saw potentially cypress's focus are taken over, so would the typing delay here help as well (like on line 86)

amsiglan: There we have delay because it is a combo-box and we wait for the dropdown to show up, where as here it is an input box, so we should not need a delay

amsiglan: Though I will take a second pass later to ensure we have focus calls on all the input elements to make the tests more robust

kavilla (approved):

PR #866, opened 2023-09-15
[Alerting 2.x] Override element covered check for name input in Composite monitor creation

amsiglan:
### Description
We are seeing flaky behavior when creating composite monitor. The name input element is sometimes not in view which prevents cypress from typing the name. Since this behavior won't happen when a user is creating the monitor it is safe to force the typing in the name element and the monitor selector elements.

### Check List

### DIFF
--- a/cypress/integration/plugins/alerting-dashboards-plugin/composite_level_monitor_spec.js
+++ b/cypress/integration/plugins/alerting-dashboards-plugin/composite_level_monitor_spec.js
@@ -71,16 +71,20 @@ describe('CompositeLevelMonitor', () => {
       cy.get('[data-test-subj="visualEditorRadioCard"]').click({ force: true });
 
       // Wait for input to load and then type in the monitor name
-      cy.get('input[name="name"]').type(SAMPLE_VISUAL_EDITOR_MONITOR);
+      cy.get('input[name="name"]').type(SAMPLE_VISUAL_EDITOR_MONITOR, {
+        force: true,
+      });
 
       // Select associated monitors
       cy.get('[data-test-subj="monitors_list_0"]').type('monitorOne', {
         delay: 50,
+        force: true,
       });
       cy.get('[title="monitorOne"]').click({ force: true });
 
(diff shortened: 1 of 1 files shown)


ruanyl:
Should this ported to `main`?

cypress/integration/plugins/alerting-dashboards-plugin/composite_level_monitor_spec.js
```diff
@@ -71,26 +71,30 @@ describe('CompositeLevelMonitor', () => {
...
 
       // Wait for input to load and then type in the monitor name
-      cy.get('input[name="name"]').type(SAMPLE_VISUAL_EDITOR_MONITOR);
+      cy.get('input[name="name"]').type(SAMPLE_VISUAL_EDITOR_MONITOR, {
+        force: true,
+      });
 
       // Select associated monitors
-      cy.get('[data-test-subj="monitors_list_0"]').type('monitorOne', {
+      cy.get('[id="associatedMonitorsList_0"]').type('monitorOne', {
```
ruanyl, line 79: Q: I'm curious why change from test id to element id? How will that mitigate the flaky test?

amsiglan: This change is for - https://github.com/opensearch-project/opensearch-dashboards-functional-test/pull/872
Basically, we should type inside a valid input element. The test id was selecting the parent div, so changed it to use the id which selects the inner input element

amsiglan:
> Should this ported to `main`?

Created PR #880 to update all alerting tests in main

cypress/integration/plugins/alerting-dashboards-plugin/composite_level_monitor_spec.js
kavilla: ah in the future, @ruanyl has a great point. You can still get the data subject and then get the id contained with that data test subject.

In this situation is seems like `[id]` is more or less in the structure of a data test subj? Perhaps to keep things similar throughout the project covert that id to data-test-subj?

kavilla (approved):

PR #874, opened 2023-09-18
[Index-mgmt 2.x] Renamed index to avoid collision for other tests

amsiglan:
### Description
During one of the tests for indices, we create an index with name `reindex_opensearch_dashboards_sample_data_ecommerce` that is superstring for the sample index `opensearch_dashboards_sample_data_ecommerce`. This new index is supposed to get deleted but sometimes it might not get deleted due to failure in backend and after that if some other test looks for the sample index and tries to select that index from a combo box, it faces collision and might end up not selecting any index. This PR renames the new index to avoid this situation.

Manual backport for 2.10 in #873 

### Issues Resolved
Fixes flaky test in Index Management Dashboards - https://github.com/opensearch-project/index-management-dashboards-plugin/issues/880

### Check List

### DIFF
--- a/cypress/integration/plugins/index-management-dashboards-plugin/indices_spec.js
+++ b/cypress/integration/plugins/index-management-dashboards-plugin/indices_spec.js
@@ -256,8 +256,7 @@ describe('Indices', () => {
   });
 
   describe('can search with reindex & recovery status', () => {
-    const reindexedIndex =
-      'reindex_opensearch_dashboards_sample_data_ecommerce';
+    const reindexedIndex = 'reindex_sample_data_ecommerce';
     const splittedIndex = 'split_opensearch_dashboards_sample_data_logs';
     before(() => {
       cy.deleteAllIndices();
--- a/cypress/integration/plugins/index-management-dashboards-plugin/indices_spec.js
+++ b/cypress/integration/plugins/index-management-dashboards-plugin/indices_spec.js
@@ -256,7 +256,7 @@ describe('Indices', () => {
   });
 
   describe('can search with reindex & recovery status', () => {
-    const reindexedIndex = `${Date.now()}_reindex_sample_data_ecommerce_`;
+    const reindexedIndex = `${Date.now()}_reindex_sample_data_ecommerce`;
     const splittedIndex = 'split_opensearch_dashboards_sample_data_logs';
     before(() => {
       cy.deleteAllIndices();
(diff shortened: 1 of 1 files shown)


kavilla:
should it be in main?

Hailong-am (approved):

cypress/integration/plugins/index-management-dashboards-plugin/indices_spec.js
```diff
@@ -256,8 +256,7 @@ describe('Indices', () => {
   });
 
   describe('can search with reindex & recovery status', () => {
-    const reindexedIndex =
-      'reindex_opensearch_dashboards_sample_data_ecommerce';
+    const reindexedIndex = 'reindex_sample_data_ecommerce';
```
ruanyl, line 259: Can we randomize the name to avoid such cases in the future?

amsiglan: This name is random enough I think.

amsiglan:
> should it be in main?

Yes, created PR #879

cypress/integration/plugins/index-management-dashboards-plugin/indices_spec.js
ruanyl: By randomize it, I mean something like `reindex_sample_data_ecommerce_${Date.now()}`. This gives you extra confidence on the uniqueness of the index name.

ruanyl:
> > should it be in main?
> 
> Yes, created PR #879

btw, you could leverage the backport workflow to port the changes to `main`

ruanyl (approved):

cypress/integration/plugins/index-management-dashboards-plugin/indices_spec.js
kavilla: > By randomize it, I mean something like `reindex_sample_data_ecommerce_${Date.now()}`. This gives you extra confidence on the uniqueness of the index name.

+1, or prefix it with `ism-` at least since re-indexing is a pretty common operation and the sample data is shared. It's probably unlikely someone will have the same name in the near future but if in a couple years it can be obfuscating something.

amsiglan: Sorry, it took me some time to get to this.
Added the date.now as prefix since the suffix won't address the issue of there was some index starting with same name.

PR #881, opened 2023-09-19
Ignore flaky test on customImportMaps when security enabled

junqiu-lei:
### Description

Ignore the flaky test on customImportMaps when security enabled

### Issues Resolved

### Check List

### DIFF
--- a/cypress/integration/plugins/custom-import-map-dashboards/documentsLayer.spec.js
+++ b/cypress/integration/plugins/custom-import-map-dashboards/documentsLayer.spec.js
@@ -5,75 +5,83 @@
 
 import { BASE_PATH } from '../../../utils/constants';
 
-describe('Documents layer', () => {
-  before(() => {
-    cy.visit(`${BASE_PATH}/app/home#/tutorial_directory/sampleData`, {
-      retryOnStatusCodeFailure: true,
-      timeout: 60000,
+if (!Cypress.env('SECURITY_ENABLED')) {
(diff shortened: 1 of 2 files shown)


gaiksaya (approved):

bbarani (approved):

cypress/integration/plugins/custom-import-map-dashboards/documentsLayer.spec.js
```diff
@@ -1,79 +0,0 @@
-/*
- * Copyright OpenSearch Contributors
- * SPDX-License-Identifier: Apache-2.0
- */
-
-import { BASE_PATH } from '../../../utils/constants';
-
-describe('Documents layer', () => {
```
kavilla, line 8: seems to be flaky due `with-security` enabled. Perhaps instead of removing we can do

something similar to [this](https://github.com/opensearch-project/opensearch-dashboards-functional-test/blob/main/cypress/integration/plugins/security-dashboards-plugin/aggregation_view.js#L27)

where we check 

```js
if (!Cypress.env('SECURITY_ENABLED')){
```
and wrap the entire test in it. So we can test it still without security and we should create a follow-up issue to look into it.

junqiu-lei: Thanks, let me update!

kavilla (approved):

PR #886, opened 2023-09-20
[Alerting 2.10] Excluding flaky tests

amsiglan:
### Description
Excluding flaky tests for composite monitors

### Issues Resolved
https://github.com/opensearch-project/alerting-dashboards-plugin/issues/737

### Check List

### DIFF
--- a/cypress/integration/plugins/alerting-dashboards-plugin/composite_level_monitor_spec.js
+++ b/cypress/integration/plugins/alerting-dashboards-plugin/composite_level_monitor_spec.js
@@ -66,7 +66,7 @@ describe('CompositeLevelMonitor', () => {
       });
     });
 
-    it('by visual editor', () => {
+    xit('by visual editor', () => {
       // Select visual editor for method of definition
       cy.get('[data-test-subj="visualEditorRadioCard"]').click({ force: true });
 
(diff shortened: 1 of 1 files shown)


cypress/integration/plugins/alerting-dashboards-plugin/composite_level_monitor_spec.js
```diff
@@ -66,7 +66,7 @@ describe('CompositeLevelMonitor', () => {
       });
     });
 
-    it('by visual editor', () => {
+    xit('by visual editor', () => {
```
kavilla, line 69: does x work? shouldn't be it.skip?

can also do something similar and conditional run these tests if its only flaky on `with-security`. Then can still verify on `without-security`

regardless, if you can make a follow up issue that would be great. thank you

amsiglan: Yes x also works on excluding these tests. Given that these tests are passing locally, and we are nearing the 2.10 deadline, I'm okay with excluding them from the CI pipeline run

kavilla (approved):
Checked yes it skipping locally.

PR #971, opened 2023-12-05
Add test orchestrator script to trigger remote workflow runner outside functional test repo for cypress tests

manasvinibs:
Issue details - https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5392

POC - https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5526

### Description

Adds a script which acts as a test orchestrator to trigger Github workflow runners in a plugin/component repository. With this feature, during release testing build team will be able to use this script to trigger the cypress tests hosted within plugins repositories and make sure if those workflows succeeded or not. Every plugin component can write tests in the same repository where their feature resides and do not have to paste the tests in functional test repo which is outside of the feature changes.

### Issues Resolved

https://github.com/opensearch-project/opensearch-dashboards-functional-test/issues/972

### Check List

### DIFF
--- a/.github/workflows/remote-cypress-workflow.yml
+++ b/.github/workflows/remote-cypress-workflow.yml
@@ -1,0 +1,36 @@
+name: Trigger Remote Cypress Workflow
+
+on:
+  workflow_dispatch:
+    inputs:
+      repo:
+        description: 'Name of the repository in {owner}/{repository} format'
+        required: true
+      workflow_name:
+        description: 'Name of the Github workflow yml file in the component repository'
+        required: true
+      os_url:
+        description: 'Release artifact of OpenSearch'
+        required: true
+      osd_url:
+        description: 'Release artifact of OpenSearch Dashboards'
+        required: true
+      branch_ref:
+        description: 'Test branch name or commit reference id'
+        required: true
+
+jobs:
+  trigger-cypress:
+    runs-on: ubuntu-latest
+    name: Remote Cypress Tests
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        
+      - name: Run Bash Script
+        env:
+          GITHUB_SECRET_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          ./remoteCypress.sh -r "${{ github.event.inputs.repo }}" -w "${{ github.event.inputs.workflow_name }}" -o "${{ github.event.inputs.os_url }}" -d "${{ github.event.inputs.osd_url }}" -b "${{ github.event.inputs.branch_ref }}"
+
--- a/remoteCypress.sh
+++ b/remoteCypress.sh
@@ -1,0 +1,113 @@
+#!/bin/bash
+
+set -e
+
+function usage() {
+    echo ""
+    echo "This script triggers GitHub workflow runners within the component repository which runs Cypress integration tests on a remote OpenSearch/Dashboards cluster."
+    echo "--------------------------------------------------------------------------"
+    echo "Usage: $0 [args]"
+    echo "Required arguments:"
+    echo -e "-r REPO\t, Name of the repository in {owner}/{repository} format"
+    echo -e "-w GITHUB_WORKFLOW_NAME\t, Name of the GitHub workflow file name with .yml extension that contain jobs that run Cypress tests in the component repository. For example, main.yaml"
+    echo -e "-o OS_URL\t, Release artifact of the OpenSearch"
+    echo -e "-d OSD_URL\t, Release artifact of the OpenSearch Dashboards"
+    echo -e "-b BRANCH_REF\t Test Branch name or commit reference id"
+    echo -e "-i BUILD_ID\t Release-specific build id for reference"
+    echo "--------------------------------------------------------------------------"
+}
+
+# Parse command-line arguments
+while getopts ":h:r:w:o:d:b:i:" opt; do
+  case $opt in
+    h)
+      usage
+      exit 1
+      ;;
+    r)
+      REPO="$OPTARG"
+      ;;
+    w)
+      WORKFLOW_NAME="$OPTARG"
+      ;;
+    o)
+      OS_URL="$OPTARG"
+      ;;
+    d)
+      OSD_URL="$OPTARG"
+      ;;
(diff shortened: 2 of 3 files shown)


remoteCypress.sh
```diff
@@ -0,0 +1,203 @@
...
+    local run_id
+    local status
+    local conclusion
+    local run_details
+    local workflow_runs
+    local matching_workflow
+    local polling_run_id_retries=1
+    local polling_workflow_completion_retries=1
+    local max_polling_run_id_retries=5  # Keep polling for the first 5 minutes to fetch the workflow run id till the workflow gets generated
+    local max_polling_workflow_completion_retries=12 # Set the polling window period to be 1 hour
```
kavilla, line 125: can this be another variable?

remoteCypress.sh
```diff
@@ -0,0 +1,203 @@
...
+    # Check if a matching workflow object was found
+    while [ -z "$matching_workflow" ] && [ $polling_run_id_retries -le $max_polling_run_id_retries ]; do
+        echo "Querying for the workflow run id..."
+        sleep 60  # Wait for 1 minute before polling for the workflow run_id till it gets created
+        
+        # Make a GET request to the GitHub API to get the list of workflow runs
+        workflow_runs=$(curl -s -H "Authorization: Bearer $GITHUB_TOKEN" \
+                    -H "Accept: application/vnd.github.v3+json" \
+                    -H "X-GitHub-Api-Version: 2022-11-28" \
+                    "https://api.github.com/repos/$REMOTE_REPO/actions/workflows/$WORKFLOW_NAME/runs")
```
kavilla, line 136: can be API url and then we append `dispatch`

kavilla:
i think we will need a way of telling that the security plugin is installed for specific logic by the test runner and a way to add cluster configs like: https://github.com/opensearch-project/opensearch-build/blob/537ccdbfc23be07e6de08236dce95906ea5b6719/manifests/2.10.0/opensearch-dashboards-2.10.0-test.yml#L14

.github/workflows/remote-cypress-workflow.yml
```diff
@@ -0,0 +1,36 @@
...
+    inputs:
+      repo:
+        description: 'Name of the repository in {owner}/{repository} format'
+        required: true
+      workflow_name:
+        description: 'Name of the Github workflow yml file in the component repository'
+        required: true
+      os_url:
+        description: 'Release artifact of OpenSearch'
+        required: true
```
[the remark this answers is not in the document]
manasvinibs: Yes, that can be totally optional.

remoteCypress.sh
```diff
@@ -0,0 +1,203 @@
...
+    fi
+
+    if [ $i -eq $MAX_RETRIES ]; then
+        echo "Maximum number of retries reached. Exiting."
+        exit 1
+    fi
+done
+
+# Function to check the status of the remote workflow by constantly polling the workflow-run
+check_remote_workflow_status() {
```
kavilla, line 115: nit: can be separate files and `source` them in a main script

remoteCypress.sh
```diff
@@ -0,0 +1,203 @@
...
+      usage
+      exit 1
+      ;;
+    r)
+      REMOTE_REPO="$OPTARG"
+      ;;
+    w)
+      WORKFLOW_NAME="$OPTARG"
+      ;;
+    o)
```
kavilla, line 41: do we think we have a pre-verification step? don't dispatch jobs. if OpenSearch and OpenSearch Dashboards can't start

.github/workflows/remote-cypress-workflow.yml
```diff
@@ -0,0 +1,36 @@
+name: Trigger Remote Cypress Workflow
+
+on:
+  workflow_dispatch:
+    inputs:
+      repo:
```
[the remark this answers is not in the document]
manasvinibs: Sure, we can take input in url format but then we will have to extract the `owner` and `repo` name from it as the workflow dispatch API expects it in that format https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#get-a-workflow-run.

remoteCypress.sh
```diff
@@ -0,0 +1,203 @@
...
+GITHUB_TOKEN=$GITHUB_SECRET_TOKEN
+# This is to uniquely identify each execution workflow. This ID has to be appended to the workflow_run 
+# name in the component repository yaml file for polling purpose. 
+UNIQUE_WORKFLOW_ID=$(uuidgen)
+echo "Unique Execution ID: $UNIQUE_WORKFLOW_ID"
+API_URL="https://api.github.com/repos/$REMOTE_REPO/actions/workflows/$WORKFLOW_NAME/dispatches"
+PAYLOAD="{\"ref\": \"$BRANCH_REF\",\"inputs\":{\"build_id\":\"$BUILD_ID\", \"OS_URL\":\"$OS_URL\", \"OSD_URL\":\"$OSD_URL\", \"UNIQUE_ID\":\"$UNIQUE_WORKFLOW_ID\"}}" 
+
+# Maximum number of retries for triggering the remote runner
+MAX_RETRIES=3
```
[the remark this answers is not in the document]
manasvinibs: This is for every run of the script, it triggers max 3 times the dispatch event API. Are you suggesting we can fail at first try if dispatch API errors out?

remoteCypress.sh
manasvinibs: Currently, I'm dispatching workflow event and not really dispatching jobs which are part of the workflow. I believe Opensearch and Opensearch Dashboards start setup script is part of the workflow yaml. I think if we have it as a jobs and if they cannot start which fails the job, it will terminate the orchestrator script as it is. But let me try to see if there is way to do pre-verification as you are suggesting in the follow-up iteration.

kavilla (approved):
Lgtm, it's not wired into anything yet. So this can be iterative and we can consider to flip to switch and add it to something that gets called automatically but we should do so in a way that calls the remote runner and still execute the tests here.

This way we can let people optionally migrate to the new setup if they please, still execute both ways for anyone consuming the results and not cause any breaking changes.

SuZhou-Joe (approved):

PR #981, opened 2023-12-21
Remove instances of hard admin credentials

derek-ho:
### Description

In 3.0.0 and backported to 2.12.0 release is a change where security demo configuration requires an initial admin password to be passed in on setup. This PR does that.

### Issues Resolved

### Check List


derek-ho:
Can somebody add backport 2.x label?

derek-ho:
Can somebody review my deletion of integtest.sh file? Is that unused in this repo? The idea is coming from: https://github.com/opensearch-project/opensearch-build/issues/497, in which it seems this file should have been migrated to build repo. Additionally it contains reference to admin:admin so I removed it, but can bring it back if that's wrong/is actively being used by this repo somehow.

.github/workflows/release-e2e-workflow-template-windows.yml
```diff
@@ -62,8 +62,8 @@ jobs:
           else
               echo "Keep OpenSearch Security"
               nohup ./opensearch-windows-install.bat &
-              timeout 900 bash -c 'while [[ "$(curl -o /dev/null -w ''%{http_code}'' -u admin:admin -k https://localhost:9200)" != "200" ]]; do echo sleeping 5; sleep 5; done'
-              curl -sk https://localhost:9200/_cluster/health?pretty -u admin:admin
+              timeout 900 bash -c 'while [[ "$(curl -o /dev/null -w ''%{http_code}'' -u admin:${{ env.OPENSEARCH_INITIAL_ADMIN_PASSWORD }} -k https://localhost:9200)" != "200" ]]; do echo sleeping 5; sleep 5; done'
```
Hailong-am, line 65: seems that `env.OPENSEARCH_INITIAL_ADMIN_PASSWORD` is not defined

Hailong-am (approved):

Hailong-am (approved):

integtest.sh
kavilla, line 1: whoops. this is the one jenkins executes

kavilla (changes requested):
integtest.sh

integtest.sh
derek-ho: Got it. I found this: https://github.com/opensearch-project/opensearch-build/blob/f92f571a42eecfc3279c217c5bd53922517054d0/scripts/components/functionalTestDashboards/build.sh. I will revert this deletion. @kavilla can you help me understand what/if any changes need to be made in integtest.sh in terms of the admin:admin credentials? I feel like there is some assumption about the setup of the OpenSearch backend that FTR would be running against, but I don't know where that is setup?

derek-ho: this line: https://github.com/opensearch-project/opensearch-dashboards-functional-test/blob/main/integtest.sh#L83

kavilla: yeah that line just need to replace it with `admin:urnewpassword` if that is the expected password. basically the integtest.sh script gets executed after an OpenSearch and OpenSearch Dashboards of the same version and successful builds get scaled up. Then Jenkins will clone this repo and execute integtest.sh. So as long as you change that password in integtest.sh it will be the credential passed when the tests are executed.

derek-ho: Do you have any idea about where those successful builds are generated? It seems like less work to modify that place to get spun up with `admin:admin`, so none of the component integtest.sh need to be changed.

SuZhou-Joe (approved):

SuZhou-Joe:
#982 for checking if all tests run successfully if security enabled.

SuZhou-Joe:
@derek-ho seems the OSD in main fails to launch according to #982 , do you know who we can ask to take a look?

derek-ho:
@kavilla @SuZhou-Joe can you take another look? Added logic to handle the logic of 2.12.0 release and onwards.

derek-ho:
> @derek-ho seems the OSD in main fails to launch according to #982 , do you know who we can ask to take a look?

Let me check this

SuZhou-Joe:
> @kavilla @SuZhou-Joe can you take another look? Added logic to handle the logic of 2.12.0 release and onwards.

LGTM

SuZhou-Joe:
@kavilla , could you please take a look on this PR and unblock the requested change if it looks good to you?

integtest.sh
```diff
@@ -78,13 +78,21 @@ then
...
 
+IFS='.' read -ra version_array <<< "$VERSION"
+
 if [ -z "$CREDENTIAL" ]
 then
-  CREDENTIAL="admin:admin"
-  USERNAME=`echo $CREDENTIAL | awk -F ':' '{print $1}'`
-  PASSWORD=`echo $CREDENTIAL | awk -F ':' '{print $2}'`
+  # Starting in 2.12.0, security demo configuration script requires an initial admin password
+  if (( ${version_array[0]} > 2 || (${version_array[0]} == 2 && ${version_array[1]} >= 12) )); then
```
kavilla, line 86: imo i dont think this need. I'm not positive if version is being passed I can check but ideally. This file will be checked out via branches or tags. So since this is on main and main is expected to 3.0.0. Which just need to make sure we backport this to 2.x then it's in the 2.12 release branch. And then dont backport any further.

One other thing is I'm not positive how this password gets set. So when the integration tests get ran they get executed after the cluster is started.

Originally this takes from an environment variable that needs to be set pre `opensearch-tar-install` is that correct?

derek-ho:
@kavilla can you re-review and merge if it looks good? I applied your suggestions and this seems to be blocking a few other PRs and workflows, thanks!

kavilla (approved):

Issue #972, opened 2023-12-05
Add test orchestrator script to trigger remote cypress test workflow GitHub runners


manasvinibs:
Add a script that builds the payload required to trigger remote workflows, executing Cypress tests in plugin repositories. Script takes following input params and trigger the github workflow in plugins repository. Using Github action API, constant polling of the workflow will be done to know if the execution successfully completed or not.
```
"-r REMOTE_REPO\t, Name of the repository in {owner}/{repository} format"
"-w GITHUB_WORKFLOW_NAME\t, Name of the GitHub workflow file name with .yml extension that contain jobs that run Cypress tests in the component repository. For example, main.yaml"
"-o OS_URL\t, Release artifact of the OpenSearch"
"-d OSD_URL\t, Release artifact of the OpenSearch Dashboards"
"-b BRANCH_REF\t Test Branch name or commit reference id"
"-i BUILD_ID\t Release-specific build id for reference"
```
This script returns either 0 (success) or 1 (failure) based on the status of the remote runner execution completion status by constantly polling the remote workflow for a set period of time. This can be used during release testing by build/Infra to know if the component(plugin) tests part of the release branch has succeeded or not.

PR #988, opened 2024-01-06
Add remote cypress orchestrator integration with integtest script

manasvinibs:
### Description

Adding remote cypress workflow orchestrator script integration with existing logic of integtest.sh script. This will keep the current behavior of integtest script as it is when triggered along with silently triggering the remote cypress workflows present in the plugin/components as defined in manifest json file and continue to poll them till its completion. 

Changes summary:

* Orchestrator integration with integtest script.
* Adding manifest json file to define plugin/components that gets onboarded to use orchestrator workflow. This manifest is used to capture plugin repo name, workflow name, branch that needs to be tested and any additional security related configuration that needs to be passed to the cypress tests running remote github workflows.
* In this iteration I'm adding Dashboards component to manifest but the workflow file and branch related information will be added once they are ready within Dashboards component.

### Issues Resolved
Follow-up to
https://github.com/opensearch-project/opensearch-dashboards-functional-test/issues/972

https://github.com/opensearch-project/opensearch-dashboards-functional-test/issues/991
### Check List

### DIFF
--- a/integtest.sh
+++ b/integtest.sh
@@ -19,11 +19,12 @@ function usage() {
     echo -e "-t TEST_COMPONENTS\t(OpenSearch-Dashboards reportsDashboards etc.), optional, specify test components, separate with space, else test everything."
     echo -e "-v VERSION\t, no defaults, indicates the OpenSearch version to test."
     echo -e "-o OPTION\t, no defaults, determine the TEST_TYPE value among(default, manifest) in test_finder.sh, optional."
+    echo -e "-r REMOTE_CYPRESS_ENABLED\t(true | false), defaults to true. Feature flag set to specify remote cypress orchestrator runs are enabled or not."
     echo -e "-h\tPrint this message."
     echo "--------------------------------------------------------------------------"
 }
 
-while getopts ":hb:p:s:c:t:v:o:" arg; do
+while getopts ":hb:p:s:c:t:v:o:r:" arg; do
     case $arg in
         h)
             usage
@@ -49,6 +50,9 @@ while getopts ":hb:p:s:c:t:v:o:" arg; do
             ;;
         o)
             OPTION=$OPTARG
+            ;;
+        r)
+            REMOTE_CYPRESS_ENABLED=$OPTARG
             ;;
         :)
             echo "-${OPTARG} requires an argument"
@@ -78,6 +82,11 @@ then
   SECURITY_ENABLED="true"
 fi
 
+if [ -z "$REMOTE_CYPRESS_ENABLED" ]
+then
+  REMOTE_CYPRESS_ENABLED="true"
+fi
+
 if [ -z "$CREDENTIAL" ]
 then
   # Starting in 2.12.0, security demo configuration script requires an initial admin password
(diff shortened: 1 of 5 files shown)


SuZhou-Joe:
Can we have a demo log or something that we can take a look that can tell us how will the feature be like in the future?

manasvinibs:
> Can we have a demo log or something that we can take a look that can tell us how will the feature be like in the future?

Certainly! Here is the link to my forked repository, `opensearch-dashboards-functional-test`, where I am currently executing the `integtest.sh` script through a GitHub workflow for testing purposes. 
https://github.com/manasvinibs/opensearch-dashboards-functional-test/actions/runs/7427907888/job/20214377315

The `Run Bash Script` logs display the initiation of the remote Cypress GitHub workflow and its continuous polling until completion. The existing logic within the integtest.sh script, designed for running in-house Cypress tests, now accommodates the wait for the remote workflow to finish before executing the remaining logic at the end. Consequently, in this scenario, an expected Cypress error is anticipated due to the absence of a configured Cypress server locally, a condition that occurs towards the conclusion of the process.

integtest.sh
```diff
@@ -100,17 +100,78 @@ echo -e "Test Files List:"
...
+components=$(jq -c '.components[]' "$REMOTE_MANIFEST_FILE")
+echo "Components: $components"
+
+for component in $components; do
+    echo "Processing for the component: $component"
+
+    repo=$(echo "$component" | jq -r '.["repository"]')
+    workflow_name=$(echo "$component" | jq -r '.["workflow-name"]')
+    os_url=$(echo "$component" | jq -r '.["opensearch"]')
+    osd_url=$(echo "$component" | jq -r '.["opensearch-dashboards"]')
```
SuZhou-Joe, line 133: What about using the
`https://artifacts.opensearch.org/releases/bundle/opensearch/{version_from_manifest}/opensearch-{version_from_manifest}-linux-x64.tar.gz`

`https://artifacts.opensearch.org/releases/bundle/opensearch-dashboards/{version_from_manifest}/opensearch-dashboards-{version_from_manifest}-linux-x64.tar.gz`

as fallbacks of os_url and osd_url?

remote_cypress_manifest.json
```diff
@@ -0,0 +1,31 @@
+{
+  "schema-version": "1.0",
+  "build": {
+    "name": "OpenSearch Dashboards Functional Test",
+    "version": "2.12.0"
```
SuZhou-Joe, line 5: On main the version should be 3.0.0 I think.

remote_cypress_manifest.json
```diff
@@ -0,0 +1,31 @@
...
+    }
+  },
+  "components": [
+    {
+      "name": "OpenSearch-Dashboards",
+      "repository": "opensearch-project/OpenSearch-Dashboards",
+      "workflow-name": "",
+      "ref": "",
+      "opensearch": "test",
+      "opensearch-dashboards": "test",
```
SuZhou-Joe, line 18: As this is an open PR, can we modify these dummy configs to valid ones?

integtest.sh
```diff
@@ -100,17 +100,78 @@ echo -e "Test Files List:"
...
+    echo "os_url: $os_url"
+    echo "osd_url: $osd_url"
+    echo "branch_ref: $branch_ref"
+
+    # Call the function for each component
+    run_remote_cypress "$repo" "$workflow_name" "$os_url" "$osd_url" "$branch_ref" 
+done
+
+# Wait for all processes to finish
+wait "${all_process_pids[@]}"
```
SuZhou-Joe, line 147: Can we make remote cypress tasks and FTRepo test task parallel and wait for both to finish? Or I think FTRepo's task will wait for a long time to start.

remote_cypress_manifest.json
manasvinibs: That's right. Updated.

integtest.sh
manasvinibs: Sure! I think OS and OSD url could be other platform/distribution combination as well in addition to linux-x64.tar. But I think we can default to linux-x64.tar when we don't have one defined in manifest. Updated in the latest revision.

remote_cypress_manifest.json
manasvinibs: Yes for main I have added linux-x64 artifact for now. Release team can choose to update component manifest json for other artifact distribution as needed.

integtest.sh
manasvinibs: We wanted to introduce this feature silently integrating in the background of existing integtest script runner for the first iteration to observe the orchestrator behavior. By keeping FTRepo test task run in the foreground after wait, we are not changing the behavior in terms of response code of the integtest script completion. We can make it run in background parallel to other remote cypress tasks, when we are ready to read the response codes from all the tasks integrated and send to integtest script caller (infra team jenkins job). 
I agree there will be delay in the start of the FTRepo task but this will help in to ensure better control, immediate visibility into the results, and simplified error handling. This approach aligns with our current requirements and allows us to seamlessly integrate the remote test results into subsequent processes. Let me know if you have other thoughts.

remote_cypress_manifest.json
```diff
@@ -0,0 +1,31 @@
+{
+  "schema-version": "1.0",
+  "build": {
+    "name": "OpenSearch Dashboards Functional Test",
+    "version": "2.12.0"
+  },
+  "ci": {
+    "image": {
```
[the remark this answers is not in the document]
manasvinibs: This manifest value is optional for now and serves to specify a container image where tests are run , as required by the Orchestrator caller (release workflow). It becomes relevant when we aim to ensure not only the execution of tests on specific version/artifacts but also the use of a specific image configured by the release team. The Orchestrator passes this information to the remote Cypress workflows running in the components/plugins, allowing them to align their GitHub workflows with the image provided by the Orchestrator. While not mandatory, it provides added flexibility to the Orchestrator, allowing it to dictate the container image if/when necessary.

remoteCypress.sh
```diff
@@ -53,61 +53,57 @@ while getopts ":h:r:w:o:d:b:i:" opt; do
...
-        -H "X-GitHub-Api-Version: 2022-11-28" \
-        -w "%{http_code}" \
-        "$API_URL/dispatches" -d "$PAYLOAD"
-}
-
-echo "Triggering the remote GitHub workflow for Cypress tests in the repository: $REPO"
+# Check if required arguments are provided
+if [[ -n "$REPO" && -n "$WORKFLOW_NAME" && -n "$OS_URL" && -n "$OSD_URL" && -n "$BRANCH_REF" ]]; then
+  # Accessing the secret as an environment variable using Github actions while invoking this script
+  GITHUB_TOKEN=$GITHUB_SECRET_TOKEN
```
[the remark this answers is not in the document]
manasvinibs: Yes `GITHUB_SECRET_TOKEN` is newly added env variable part of the github workflow - https://github.com/opensearch-project/opensearch-dashboards-functional-test/blob/main/.github/workflows/remote-cypress-workflow.yml#L33  but the token used is existing one within the team's repo used by other github workflows such as [delete workflow](https://github.com/opensearch-project/opensearch-dashboards-functional-test/blob/main/.github/workflows/delete_backport_branch.yml#L15)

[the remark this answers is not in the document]
ruanyl: I see, is there cases that this script is run outside of github workflow where `GITHUB_TOKEN` is not set?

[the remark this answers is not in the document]
manasvinibs: Currently, this will be run as part of `integtest` script. Along with that one can manually trigger the remoteCypress script using dispatch event. In both the cases,  I believe [github_cypress_workflow](https://github.com/opensearch-project/opensearch-dashboards-functional-test/pull/988/files#diff-34376fefedd23fbe419947eeff0721a33e59b9b9504d03cbcfcd14cb8f70c937R35) would have set the `GITHUB_SECRET_TOKEN` env variable  to `GITHUB_TOKEN` in the repo.

integtest.sh
```diff
@@ -102,17 +102,84 @@ echo -e "Test Files List:"
...
+    local workflow_name="$2"
+    local os_url="$3"
+    local osd_url="$4"
+    local branch_ref="$5"
+
+    # Call the remoteCypress.sh script with the required arguments
+    source remoteCypress.sh -r "$repo" -w "$workflow_name" -o "$os_url" -d "$osd_url" -b "$branch_ref" &
+    bg_process_pid=$!
+    echo "PID for the repo $repo is : $bg_process_pid"
+    all_process_pids+=($bg_process_pid)
```
kavilla, line 119: i think it might be worth to consider using pid files https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/config/opensearch_dashboards.yml#L122

i didn't realize this logic existed but has long as the file is created we can reference the process id file instead of it in memory

integtest.sh
```diff
@@ -102,17 +102,84 @@ echo -e "Test Files List:"
...
+}
+
+# Read inputs from the manifest file using jq
+REMOTE_MANIFEST_FILE="remote_cypress_manifest.json"
+
+# Parse the JSON file using jq and iterate over the components array
+components=$(jq -c '.components[]' "$REMOTE_MANIFEST_FILE")
```
kavilla, line 126: nit: for readability it might be helpful to have a helper function for this and perhaps it returns back a tuple with the required information

integtest.sh
```diff
@@ -102,17 +102,84 @@ echo -e "Test Files List:"
...
+    echo "Processing for the component: $component"
+
+    repo=$(echo "$component" | jq -r '.["repository"]')
+    workflow_name=$(echo "$component" | jq -r '.["workflow-name"]')
+    os_url=$(echo "$component" | jq -r '.["opensearch"]')
+    osd_url=$(echo "$component" | jq -r '.["opensearch-dashboards"]')
+    branch_ref=$(echo "$component" | jq -r '.["ref"]')
+
+    # Set default values if the opensearch and opensearch-dashboards are not set in the manifest
+    os_url=${os_url:-https://artifacts.opensearch.org/releases/bundle/opensearch/$release_version/opensearch-$release_version-linux-x64.tar.gz}
```
kavilla, line 141: `linux-x64`. i think it should get the operating system of the runner and arch

integtest.sh
```diff
@@ -102,17 +102,84 @@ echo -e "Test Files List:"
...
+    osd_url=${osd_url:-https://artifacts.opensearch.org/releases/bundle/opensearch-dashboards/$release_version/opensearch-dashboards-$release_version-linux-x64.tar.gz}
+
+    echo "repo: $repo"
+    echo "workflow_name: $workflow_name"
+    echo "os_url: $os_url"
+    echo "osd_url: $osd_url"
+    echo "branch_ref: $branch_ref"
+
+    # Call the function for each component
+    run_remote_cypress "$repo" "$workflow_name" "$os_url" "$osd_url" "$branch_ref"
```
kavilla, line 151: i think we should add a flag or env variable that enables us to conditional run the run remote cypress tests are not. since this repo is now being locked as well to commits a new build will have to be made if there are any issues. so it pre-planning a feature flag could be a fallback to disabling this if it causes issues.

integtest.sh
manasvinibs: I've adjusted the manifest to get operating system/arch and constructing the url while parsing the manifest with the base url being `https://ci.opensearch.org/ci/dbc/distribution-build/`.

integtest.sh
manasvinibs: I referenced the [developer guide doc](https://github.com/opensearch-project/opensearch-dashboards-functional-test/blob/main/DEVELOPER_GUIDE.md#experimental-features) about setting up flag for experimental features and its tests. As my feature is not actually running cypress tests but instead running orchestrator workflows, I've introduced a simple boolean flag as part of `integtest.sh` script params defaulting to `true`. Build script which triggers `integtest.sh` can set the feature flag to true/false as required. Is this the behavior you were expecting or is there some other way to configure this setting in FTRepo?

manasvinibs:
@kavilla @SuZhou-Joe @ruanyl Can you please review and let me know if there anything else you would like me to address to get this checked in? I'm hoping to get this in for 2.12 and would like to have some time to bake this changes in main and address any issues related to integration.

remote_cypress_manifest.json
```diff
@@ -0,0 +1,31 @@
...
+    }
+  },
+  "components": [
+    {
+      "name": "OpenSearch-Dashboards",
+      "repository": "opensearch-project/OpenSearch-Dashboards",
+      "workflow-name": "",
+      "ref": "",
+      "operating-system": "linux",
```
kavilla, line 17: os

integtest.sh
```diff
@@ -102,17 +111,111 @@ echo -e "Test Files List:"
...
+}
+
+# Wait for all processes to finish
+wait_file() {
+    while read -r pid; do
+        wait "$pid"
+    done < "$pid_file"
+}
+
+if [[ $REMOTE_CYPRESS_ENABLED = "true" &&  $ORCHESTRATOR_FEATURE_FLAG = 'true' ]]; then
```
kavilla, line 170: check if env variable for github token if not then don't execute this.

kavilla (approved):

integtest.sh
manasvinibs: Sure! I have a follow-up PR to add manifest data to this. I'll address this one in that.

ananzh (approved):

peterzhuamazon:
Hi @manasvinibs if you ever want to backport please note that I have add placeholder for `-r` in case 1.x/2.x wont fail when we call `-r false`.
https://github.com/opensearch-project/opensearch-dashboards-functional-test/pull/1070

We have to disable the remote test settings due to it failing out production integTest on Jenkins. Thanks.

PR #1088, opened 2024-02-16
Fix in-context reporting tests

Swiddis:
### Description

When checking for a reporting download being triggered, the tests are waiting for a loading screen. This causes test failures if the download happens faster than React renders. In one test run, the test failed when the API returned in 18 ms (whereas a passing test in the same run had a return time of 224 ms). This PR updates the tests to instead check for the success toast.

### Issues Resolved

N/A

### Check List


Swiddis:
Will also need to forward-port to 2.x and main, starting here since FTR is failing in RC generation.

derek-ho (approved):
Thanks

kavilla (approved):

PR #1120, opened 2024-02-22
[OSD] Skip cypress tests for data explorer with security enabled

kavilla:
### Description

When security is enabled the url contains the tenancy. For example, 
```
localhost:5601/app/dashboards?security_tenant=private
```

Which will fail tests that verify URLs if written without security plugin and tenancy enabled. Also with the security plugin enabled then the query string is masked. Which will make it more difficult to verify URL tests without checking the expecting elements within the app. So to mitigate this we will just check the URLs only if security is not enabled but keep the tests running if enabled.

Also includes browserPermissions as one of the tests would pop up and asks but needs permission. Also commented on the cy.visit. Depending on how the tests are executed this could be a re-navigate on origin. But since it just focuses on visiting but fails the tests when it does visit but not validate anything on the location then commented out. 

### Issues Resolved

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/5633

### Check List

### DIFF
--- a/cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/data_explorer/shared_links.spec.js
+++ b/cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/data_explorer/shared_links.spec.js
@@ -52,51 +52,58 @@ describe('shared links', () => {
     cy.waitForSearch();
   });
 
-  describe('shared links with state in query', () => {
-    it('should allow for copying the snapshot URL', function () {
-      const url = `http://localhost:5601/app/data-explorer/discover#/?_a=(discover:(columns:!(_source),isDirty:!f,sort:!()),metadata:(indexPattern:'logstash-*',view:discover))&_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:'2015-09-19T13:31:44.000Z',to:'2015-09-24T01:31:44.000Z'))&_q=(filters:!(),query:(language:kuery,query:''))`;
-      cy.getElementByTestId('shareTopNavButton').should('be.visible').click();
-      cy.getElementByTestId('copyShareUrlButton')
-        .invoke('attr', 'data-share-url')
-        .should('eq', url)
-        .then((url) => {
-          cy.log(url);
-          cy.request(url).its('status').should('eq', 200);
-        });
+  if (!Cypress.env('SECURITY_ENABLED')) {
+    describe('shared links with state in query', () => {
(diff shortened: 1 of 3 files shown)


cypress/integration/core-opensearch-dashboards/opensearch-dashboards/dashboard_share_copy_link_test.js
```diff
@@ -28,8 +28,9 @@ if (Cypress.env('SECURITY_ENABLED')) {
             // Assert that the clipboard has data
             expect(clipboardData).to.have.length.greaterThan(0);
 
-            cy.visit(clipboardData);
-            cy.waitForLoader();
+            // TODO: tries to re-route with a different origin
+            // cy.visit(clipboardData);
```
kavilla, line 32: @leanneeliatra would you think this is ok?

abbyhu2000 (approved):

kavilla:
> As 2.12 has been released, I will remove the `backport 2.12` label.

In case there is a patch. Do we want to backport to 2.12 since those tests are inaccurate when installing the security plugin.

ashwin-pc (approved):

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/data_explorer/shared_links.spec.js
```diff
@@ -52,51 +52,58 @@ describe('shared links', () => {
...
-    });
-
-    it('should allow for copying the saved object URL', function () {
-      const url =
-        'http://localhost:5601/app/data-explorer/discover/#/view/ab12e3c0-f231-11e6-9486-733b1ac9221a?_g=%28filters%3A%21%28%29%2CrefreshInterval%3A%28pause%3A%21t%2Cvalue%3A0%29%2Ctime%3A%28from%3A%272015-09-19T13%3A31%3A44.000Z%27%2Cto%3A%272015-09-24T01%3A31%3A44.000Z%27%29%29';
-
-      cy.getElementByTestId('exportAsSavedObject')
-        .get('.euiRadio__input')
-        .should('be.disabled');
+  if (!Cypress.env('SECURITY_ENABLED')) {
```
AMoo-Miki, line 55: I am not a fan of this.

We could just strip the tenant from the query string and then compare. In fact, checking if the URL is identical is not a good idea; we should check if the elements we want are there, in whatever order.

kavilla:
Closing. Will track with this: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/6220#issuecomment-2031290984

PR #1128, opened 2024-02-26
Add tests for shared links when multi-tenancy is enabled

cwperks:
### Description

Adding tests to go with corresponding PR in the security-dashboards-plugin repo: https://github.com/opensearch-project/security-dashboards-plugin/pull/1804

These tests ensure that the `security_tenant` url param is added onto shared links - both shortened URLs and normal.

### Issues Resolved

- https://github.com/opensearch-project/security-dashboards-plugin/issues/1769

### Check List

### DIFF
--- a/cypress/integration/plugins/security-dashboards-plugin/multitenancy_shared_links.js
+++ b/cypress/integration/plugins/security-dashboards-plugin/multitenancy_shared_links.js
@@ -1,0 +1,55 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { STACK_MANAGEMENT_PATH } from '../../../utils/dashboards/constants';
+
+import { switchTenantTo } from './switch_tenant';
+import 'cypress-real-events/support';
+
+Cypress.automation('remote:debugger:protocol', {
+  command: 'Browser.grantPermissions',
+  params: {
+    permissions: ['clipboardReadWrite', 'clipboardSanitizedWrite'],
+    origin: window.location.origin,
+  },
+});
+
+if (Cypress.env('SECURITY_ENABLED')) {
+  describe('Multi Tenancy Tests for Shared Links: ', () => {
+    it('Tests to ensure that copy link includes security tenant as url param', () => {
+      cy.visit(STACK_MANAGEMENT_PATH);
+      cy.waitForLoader();
+      switchTenantTo('private');
+      cy.getElementByTestId('toggleNavButton').click();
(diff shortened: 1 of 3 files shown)


wbeckler:
Rather than adding tests here, you can add tests inside your original feature PR and invoke the FT Repo as a test runner.

cwperks:
@wbeckler Thank you for linking to the PR and having a sample to reference. I would like to use that eventually and have more cypress tests in the security repo, but I think there will be a little more work that needs to be done in OSD core first to run cypress tests with different plugins installed (such as the security plugin). I will look into what is required in OSD core to trigger the FTR repo to run w/ security and file an issue in the OSD core repo. 

wdyt about keeping this PR in this repo until OSD core supports triggering tests w/ security?

cypress/integration/plugins/security-dashboards-plugin/multitenancy_shared_links.js
```diff
@@ -0,0 +1,55 @@
...
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { STACK_MANAGEMENT_PATH } from '../../../utils/dashboards/constants';
+
+import { switchTenantTo } from './switch_tenant';
+import 'cypress-real-events/support';
+
+Cypress.automation('remote:debugger:protocol', {
```
kavilla, line 11: Can't we in the cypress.json add:
```
    "browserPermissions": {
      "clipboard": "allow"
    }
```

That way other tests can just assume it can use the clipboard.

kavilla (approved):
I agree with both you and @wbeckler. Perhaps do you mind keeping track of the tests that you add in this repo in an issue within your repo. In that way, we will merge it here and then if there is ever a shift in the testing CI then you know which tests that can be targeted first.

cypress/integration/plugins/security-dashboards-plugin/multitenancy_shared_links.js
DarshitChanpura: +1

DarshitChanpura (commented):
Thanks @cwperks for raising this. This file should be deleted: https://github.com/opensearch-project/opensearch-dashboards-functional-test/blob/main/cypress/integration/core-opensearch-dashboards/opensearch-dashboards/dashboard_share_copy_link_test.js in lieu of test introduced in this PR.

SuZhou-Joe:
No update for 6 months and rerun the CI still failed, close the PR. Please feel free to open if any update.

PR #1270, opened 2024-05-06
Add further creation waits to notebooks tests

Swiddis:
### Description

Since notebooks are still flaky on notebook creation, even with refresh logic added (see #1250), I think the reason the tests are still failing is because notebook creation is flaky in some index refresh race condition scenarios. This PR refactors the notebook creation to drop the refresh attempt and only check for a successful creation message, on the assumption that notebooks should have a functional requirement to always show a valid notebook after stating a successful creation.

I'm not sure this fixes the flaky tests in CI (still struggling to understand the root cause and get exact repro steps), but in local testing this passed 10 consecutive runs (the old version failed after 6 runs and a version with both parts fails almost every run). This should also pass the ball out of the testing repo's court, I've created a corresponding issue: https://github.com/opensearch-project/dashboards-observability/issues/1822. 

Also needs backport to 2.x and main.

### Issues Resolved

N/A

### Check List

[^1]: We're still keeping the refresh, see below comment.

### DIFF
--- a/cypress/integration/plugins/observability-dashboards/6_notebooks.spec.js
+++ b/cypress/integration/plugins/observability-dashboards/6_notebooks.spec.js
@@ -10,7 +10,6 @@ import {
   BASE_PATH,
   delayTime,
   MARKDOWN_TEXT,
-  OBSERVABILITY_INDEX_NAME,
 } from '../../../utils/constants';
 
 import { skipOn } from '@cypress/skip-test';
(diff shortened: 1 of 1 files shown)


Swiddis:
Nvm, looks like there's still flakiness after this change -- started being able to reproduce the notebook failures by running the tests on a cluster with many nodes.

joshuali925 (approved):

ashwin-pc (approved):

Hailong-am (approved):

PR #1266, opened 2024-05-03
Fix paths for MDS api changes

derek-ho:
### Description

After introducing MDS feature in 2.14, security dashboards API changed to include a query param - dataSourceId= to indicate which datasource the call is being made for. This was not being appropriately handled in the intercept, leading to timeout. This PR adds that param to the intercept calls so that tests work appropriately. 

### Issues Resolved

[List any issues this PR will resolve]

### Check List

- [ ] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1271, opened 2024-05-06
Add * at the end of mocked API calls to support MDS api changes

derek-ho:
### Description
Add * at the end of mocked API calls to support MDS api changes. After security dashboards plugin added support for MDS, API endpoint changed because we need to add dataSourceId as a query param. This results in the mocked API calls not being triggered and the tests timing out. This adds a * at the end of the paths to have the real API calls trigger the mocked commands. 

### Issues Resolved

Forward port of: #1266 

### Check List


kavilla (approved):

SuZhou-Joe:
#1276 Seems this PR does the same change to the intercept.

PR #1321, opened 2024-05-10
Fix: test failures related to saved objects(tenant)

SuZhou-Joe:
### Description

This is a PR to mainly address the bugs described as below:

### Bug 1: TenantInfo Mismatch in Requests and Tests

Several test cases concerning saved objects, particularly those within the core-dashboards, are affected by this issue.

The problem arises from a discrepancy between the `tenantInfo` specified in the request and the `tenantInfo` expected within the test. In code snippets like:

```javascript
CURRENT_TENANT.newTenant = 'global';
```

The Cypress intercepts `visit` commands and certain commands related to saved objects to inject `tenantInfo` into the request. However, some calls bypass this interception, such as data ingestion operations directly sent to the OpenSearch cluster. Consequently, data is injected into the `private` tenant instead of the expected `global` tenant while the page is being visited, leading to test failures.

### Bug 2: TenantInfo Clearance During Test Suite Switching

Cypress employs a mechanism whereby the cookie containing `tenantInfo` is cleared when switching between test suites. This mechanism is causing test failures because it fails to refresh tenant settings before executing assertions and tests.

This pull request aims to address the aforementioned test failures. After merging this PR, it is expected that the test failures will mirror those occurring when security features are disabled.

And you may find there are many places fleshing tenant settings both in `before` and `beforeEach`, it is because that `before` is executed before `beforeEach` and in order to keep tenantInfo consistent throughout the whole test suites, we had to do that.

### Issues Resolved

### Check List

### DIFF
--- a/cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/data_explorer/discover.spec.js
+++ b/cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/data_explorer/discover.spec.js
@@ -148,11 +148,6 @@ describe('discover app', { scrollBehavior: false }, () => {
         cy.setTopNavDate(fromTime, toTime);
       });
 
-      beforeEach(() => {
-        CURRENT_TENANT.newTenant = 'global';
-        cy.fleshTenantSettings();
-      });
-
       it('should show "no results"', () => {
         cy.getElementByTestId('discoverNoResults').should('be.exist');
       });
--- a/cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/data_explorer/discover.spec.js
+++ b/cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/data_explorer/discover.spec.js
@@ -142,8 +142,15 @@ describe('discover app', { scrollBehavior: false }, () => {
       const toTime = 'Jun 12, 1999 @ 11:21:04.000';
 
       before(() => {
+        CURRENT_TENANT.newTenant = 'global';
+        cy.fleshTenantSettings();
         cy.switchDiscoverTable('new');
         cy.setTopNavDate(fromTime, toTime);
+      });
+
+      beforeEach(() => {
+        CURRENT_TENANT.newTenant = 'global';
+        cy.fleshTenantSettings();
       });
 
       it('should show "no results"', () => {
--- a/cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/data_explorer/discover.spec.js
+++ b/cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/data_explorer/discover.spec.js
@@ -27,17 +27,7 @@ const indexSet = [
 describe('discover app', { scrollBehavior: false }, () => {
   before(() => {
     CURRENT_TENANT.newTenant = 'global';
-    if (Cypress.env('SECURITY_ENABLED')) {
-      /**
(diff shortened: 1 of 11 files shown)


cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/data_explorer/discover_advanced_settings.spec.js
```diff
@@ -221,7 +235,13 @@ describe('discover_advanced_setting', () => {
   });
 
   describe('Number of terms advanced setting', () => {
+    beforeEach(() => {
```
wanglam, line 238: Nit: not a big concern, shall we move `beforeEach` to outermost describe? It seems beforeEach be added to each inside describe.

wanglam (approved):

wanglam (commented):
Seems [the ci](https://github.com/opensearch-project/opensearch-dashboards-functional-test/actions/runs/9030473592/job/24814817875?pr=1321) failed in without security cases. Shall we fix this in this PR?

raintygao (approved):

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/data_explorer/discover_advanced_settings.spec.js
SuZhou-Joe: Oh, I thought beforeEach only take effect inside a describe, if that's the case then I will sure to move to outermost describe.

SuZhou-Joe:
> Seems there are 4 failed cases in [the ci](https://github.com/opensearch-project/opensearch-dashboards-functional-test/actions/runs/9030473592/job/24814818095?pr=1321) results. Shall we fix this in this PR?

I'd say no because that is not related to tenant and for now I do not have idea on how to fix those test.

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/data_explorer/discover_advanced_settings.spec.js
SuZhou-Joe: Thanks for calling it out, I learned a lot from this comment.

cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/data_explorer/discover.spec.js
```diff
@@ -27,17 +27,7 @@ const indexSet = [
...
-    if (Cypress.env('SECURITY_ENABLED')) {
-      /**
-       * Security plugin is using private tenant as default.
-       * So here we'd need to set global tenant as default manually.
-       */
-      cy.changeDefaultTenant({
-        multitenancy_enabled: true,
-        private_tenant_enabled: true,
-        default_tenant: 'global',
-      });
```
Hailong-am, line 39: if we set default tenant as `global` in security system index and not specify tenant for each visit, would that solve the tenant issue? 

is there a case need tenant to be `private`? if there is no such case, i think we can set default tenant as `global` instead `private`

SuZhou-Joe: For dashboard, no I suppose. And I agree that we should set default tenant as `global` to hide the complicated tenant logic.

Issue #1323, opened 2024-05-10
[BUG] Visbuilder dashboard.spec.js failed


SuZhou-Joe:
### What is the bug?

https://github.com/opensearch-project/opensearch-dashboards-functional-test/assets/13493605/484a3be8-e7b3-468e-9d23-e7223f489292

### How can one reproduce the bug?
_Steps to reproduce the behavior._

### What is the expected behavior?
_A clear and concise description of what you expected to happen._

### What is your host/environment?
_Operating system, version._

### Do you have any screenshots?
_If applicable, add screenshots to help explain your problem._

### Do you have any additional context?
_Add any other context about the problem._

dblock:
@SuZhou-Joe was this fixed with #1371? Can it be closed?

Catch All Triage - [1](https://github.com/dblock) [2](https://github.com/dbwiddis) [3](https://github.com/krisfreedain) [4](https://github.com/akolarkunnu) [5](https://github.com/mingshl) [6](https://github.com/getsaurabh02)

SuZhou-Joe:
Sure, close that.

Issue #1322, opened 2024-05-10
[BUG] Visbuilder basic.spec.js failed


SuZhou-Joe:
### What is the bug?

The UI component is overlapped and it seems the interaction is not correct as after clicking the button, the axis form disappeared.

### Overlapped
<img width="247" alt="image" src="https://github.com/opensearch-project/opensearch-dashboards-functional-test/assets/13493605/80bc21d9-1c3a-4578-9898-5f86611022ca">

### Disappeared
<img width="284" alt="image" src="https://github.com/opensearch-project/opensearch-dashboards-functional-test/assets/13493605/1c3ba5c9-eabf-43b4-839b-5d1d1991be84">

### The video

https://github.com/opensearch-project/opensearch-dashboards-functional-test/assets/13493605/4a65c754-c6e4-4a7f-9a57-4cf512ef8133

### How can one reproduce the bug?
_Steps to reproduce the behavior._

### What is the expected behavior?
_A clear and concise description of what you expected to happen._

### What is your host/environment?
_Operating system, version._

### Do you have any screenshots?
_If applicable, add screenshots to help explain your problem._

### Do you have any additional context?
_Add any other context about the problem._

dblock:
@SuZhou-Joe was this fixed with #1371? Can it be closed?

Catch All Triage - [1](https://github.com/dblock) [2](https://github.com/dbwiddis) [3](https://github.com/krisfreedain) [4](https://github.com/akolarkunnu) [5](https://github.com/mingshl) [6](https://github.com/getsaurabh02)

SuZhou-Joe:
Nope, in main it still failed. @ashwin-pc Could you please take a look? This failure happens only in main branch.

Issue #1298, opened 2024-05-08
[FLAKY TEST] Share Copy link Test is flaky


derek-ho:
### What is the bug?
Share copy link test: https://github.com/opensearch-project/opensearch-dashboards-functional-test/blob/main/cypress/integration/core-opensearch-dashboards/opensearch-dashboards/dashboard_share_copy_link_test.js is flaky and failed 2.14 release. We should look into it and see how we can change the test to make it not fail. 
### How can one reproduce the bug?
_Steps to reproduce the behavior._

### What is the expected behavior?
_A clear and concise description of what you expected to happen._

### What is your host/environment?
_Operating system, version._

### Do you have any screenshots?
_If applicable, add screenshots to help explain your problem._

### Do you have any additional context?
_Add any other context about the problem._

derek-ho:
@cwperks can you assign both you and I to look into this before the next release?

PR #1352, opened 2024-06-04
[Infra] Add CI groups for OSD core test cases to avoid flaky test

SuZhou-Joe:
### Description

In the release of 2.14, OSD core has 3 flaky tests listed in `Issues Resolved` section.  This PR splits the test cases into CI groups to make those cases pass.

### Issues Resolved

https://github.com/opensearch-project/opensearch-dashboards-functional-test/issues/1323
https://github.com/opensearch-project/opensearch-dashboards-functional-test/issues/1322
https://github.com/opensearch-project/opensearch-dashboards-functional-test/issues/1298

### Check List


ruanyl (approved):

rishabh6788:
Before we merge this PR can you please help answer below questions?
1. If the reasoning behind splitting into groups is resource constraint or data cleanup, then:
  a. Did you check running with bigger cluster with appropriate heap size?
  b. If it is data clean up issue then is there any root cause analysis for it?
2. Is there any logic behind grouping or is it random?
3. It will be helpful if you could elaborate on the logic used to group the tests. 
4. Please move the grouping logic outside of workflow yml file, so that the same can be used by CI integ test logic.

peterzhuamazon:
Add up to the last point raised by @rishabh6788 , we should consider using the existing `test_finder` and `integtest` shell script to apply the logic.

Our Python code can then apply these logics with parameters and onboard to Jenkins with smaller effort.

Thanks.

SuZhou-Joe:
1. We have not get the bandwidth to find the root cause and splitting into CI groups is the quick solution to unblock the release. And from @kavilla 's observation, it has something to do with the data clean up.
2/3 I feel like the grouping logic is random, but @kavilla @ashwin-pc please correct me if I was wrong.
4. I do not think it is feasible to make the logic reuseable as most of the logic inside yml is to arrange a matrix strategy so that all the test cases can be ran parallel. For jenkins there will be a different syntax to support matrix I suppose. @peterzhuamazon any advice on making the code inside yml file reuseable?

gaiksaya:
> 1. We did not get the bandwidth to find the root cause and splitting into CI groups is the quick solution to unblock the release. And from @kavilla 's observation, it has something to do with the data clean up.
>        2/3. I feel like the grouping logic is random, but @kavilla @ashwin-pc please correct me if I was wrong.

Can we invest sometime in root causing before jumping to quick solutions and conclusions? I understand the timeline concern but changes on the infra side also won't be as straight forward. Need to plan it properly not to break 1x line compatibility and also accommodate other components. Wondering if major concern here is data clean or CPU/resources constraints or flaky tests.

SuZhou-Joe:
1. From @kavilla 's observation, it has something to do with the data clean up. Addressing the root cause will be a following item.

2/3. I feel like the grouping logic is random, but @kavilla @ashwin-pc please correct me if I was wrong.

4. Yeah, the logic to format and get the specs under specific CI groups can be reused. Will optimize that.

rishabh6788:
@ruanyl We were still waiting on getting our questions answered and proposed a few changes. Any reason why this was merged before getting those questions answered?
@SuZhou-Joe @getsaurabh02 @peterzhuamazon

getsaurabh02:
If this PR was to fix the flaky test case failures through grouping, CI for this PR should not have failed? 😖

ruanyl:
@rishabh6788 My bad, I wasn't aware of the open questions. Shall we revert the PR?

gaiksaya:
> If this PR was to fix the flaky test case failures through grouping, CI for this PR should not have failed? 😖
> 
> @ashwin-pc should we root cause why is CI still failing post grouping : https://github.com/opensearch-project/opensearch-dashboards-functional-test/actions/runs/9410026571/job/25921007182?pr=1368

To add to this, are we sure that the test cases are exactly same between OSD repo and FT repo?

SuZhou-Joe:
@ruanyl Never mind, it won't affect the existing infra pipeline and after merging the code to 2.x, and we could ask other PRs to ensure all the flows pass before merging into 2.x.

peterzhuamazon:
> If this PR was to fix the flaky test case failures through grouping, CI for this PR should not have failed? 😖
> 
> @ashwin-pc should we root cause why is CI still failing post grouping : https://github.com/opensearch-project/opensearch-dashboards-functional-test/actions/runs/9410026571/job/25921007182?pr=1368

We need to understand if this is a code / resource / cleanup issues and set out a plan to fix it permanently if possible.

kavilla:
Sorry about the delay.

The grouping with OSD cigroups weren't random. It split out the Discover the tests because each group is could be really slow. It also somewhat ensured that the ciGroups that are ran together would avoid causing failures in another group. We attempted to avoid couple tests in the same group that used another tests datasets.

Then we added each other test were we see fit (somewhat related) and avoiding the Discover tests.

For the resource clean up we see this problem in hosted environments anyways where then users have to force a refresh. Taking a random repo issue: https://github.com/opensearch-project/notifications/issues/790

We mark things for deletion but that's not guaranteeing that these items are actually being deleted in the cluster. Then we move on to the next expensive test that inserts more data and tries to query that data and potentially a search phase execution error occurs because JVM has too much pressure.

But refreshing after every delete is also a very expensive process so there could be runtime issues.

So I'm really not sure of a better solution then to make sure to use ciGroups at least for OpenSearch Dashboards.

OpenSearch Dashboards on main has 58 plugins alone. When we forked the selenium tests required they be ran in CiGroups and we have plans to migrate to our selenium tests to Cypress. When we migrated most of the Discover selenium tests to cypress we were easily able to choke the cluster with the amount of times we write a bunch of data and delete unless we are allocated a production-like cluster.

PR #1377, opened 2024-06-11
Fix Flaky Discover Table Tests

LDrago27:
### Description

Fix Flaky tests in Discover Table.

Without Security

https://github.com/opensearch-project/opensearch-dashboards-functional-test/assets/38322563/e800b3fd-989e-444e-b5e4-eec8b1e62c1f

With Security

https://github.com/opensearch-project/opensearch-dashboards-functional-test/assets/38322563/bc377c53-ce91-4e42-903c-4bbb2e34bdcf

### Check List

### DIFF
--- a/cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/data_explorer/discover_table.spec.js
+++ b/cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/data_explorer/discover_table.spec.js
@@ -408,22 +408,27 @@ describe('discover_table', () => {
   describe('AutoSize table', () => {
     describe('Legacy Table', () => {
       it('check table Auto Size with change in time range', function () {
-        cy.get('[aria-label="Next"]')
+        cy.get('[data-test-subj="docTableExpandToggleColumn"]')
           .its('length')
           .then((noEntries) => {
             cy.setTopNavDate(
               'Sep 22, 2015 @ 14:00:00.000',
-              'Sep 22, 2015 @ 18:00:00.000'
+              'Sep 22, 2015 @ 14:05:00.000'
             );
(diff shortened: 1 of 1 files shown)


cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/data_explorer/discover_table.spec.js
```diff
@@ -459,9 +464,9 @@ describe('discover_table', () => {
           .then((noEntries) => {
             cy.setTopNavDate(
               'Sep 22, 2015 @ 14:00:00.000',
-              'Sep 22, 2015 @ 18:00:00.000'
+              'Sep 22, 2015 @ 14:30:00.000'
             );
-            cy.waitForLoader();
```
ashwin-pc, line 464: why did you change this? wait for loader is more reliable since it waits for a specific element and not just the an arbitrary time which can cause the test to become flakey

LDrago27: While running the tests, I found that waiting for page load didn't make sense in few occasions. While applying the filter to a table the entire page doesn't reload. In those scenarios instead of waiting for the number of entires to change , cypress picks up the older result and the evaluation condition fails. 
To avoid such scenarios I have put a wait time.

LDrago27: @ashwin-pc Updated the PR to remove waits, instead added verification for hits which has resolved the previous issues.

ashwin-pc (approved):

ananzh (approved):

PR #1408, opened 2024-06-13
feat: freeze browser version to 112

SuZhou-Joe:
### Description

Github CI is using the latest chromium to run the Cypress test, which leads to forever hang in some test cases.
According to [#4241](https://github.com/opensearch-project/opensearch-build/issues/4241) , we will freeze the Chromium version to 112, exactly same as what Jenkins CI uses.

### Issues Resolved

### Check List

### DIFF
--- a/.github/workflows/release-e2e-workflow-template.yml
+++ b/.github/workflows/release-e2e-workflow-template.yml
@@ -46,6 +46,7 @@ jobs:
       - name: Download Chromium 112
         working-directory: cypress-test
         run: |
+          ## using Chromium 112 for stable test env, for more detail: https://github.com/opensearch-project/opensearch-build/issues/4241
           source ./browser_downloader.sh
           download_chromium
           sudo ln -s $chromium_path /usr/bin/chromium
--- a/.github/workflows/release-e2e-workflow-template.yml
+++ b/.github/workflows/release-e2e-workflow-template.yml
@@ -143,7 +143,6 @@ jobs:
           # remove success video to save space and memory
           CYPRESS_DELETE_SUCCESS_VIDEOS: true
         with:
-          numTestsKeptInMemory: 1
           working-directory: cypress-test
           command: ${{ inputs.test-command }}
           wait-on: 'http://localhost:5601'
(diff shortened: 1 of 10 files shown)


.github/workflows/release-e2e-workflow-template.yml
```diff
@@ -127,7 +139,11 @@ jobs:
       - run: npx cypress cache path
       - name: Cypress tests
         uses: cypress-io/github-action@v2
+        env:
+          # remove success video to save space and memory
+          CYPRESS_DELETE_SUCCESS_VIDEOS: true
```
peterzhuamazon, line 144: Instead of record then delete videos, just disable recording:
https://github.com/opensearch-project/opensearch-dashboards-functional-test/blob/main/integtest.sh#L219-L223

peterzhuamazon (approved):

.github/workflows/release-e2e-workflow-template.yml
SuZhou-Joe: Yes, but we want to keep the video if the specs failed.

cypress/support/index.js
```diff
@@ -19,7 +19,7 @@
 // ***********************************************************
 
 // Import commands.js using ES2015 syntax:
-import '../utils/commands';
+import { currentBackendEndpoint } from '../utils/commands';
```
ruanyl, line 22: Unrelated?

ruanyl (approved):

.github/workflows/release-e2e-workflow-template.yml
SuZhou-Joe: And even if we downgrade chromium version to 112, the security test will still fail. I am trying to reduce the memory our test cases are using.

cypress/support/index.js
SuZhou-Joe: There is a lint warning in this file that we import `'../utils/commands'` twice, so I combine the two import into one.

SuZhou-Joe:
Convert to draft as it is still WIP.

.github/workflows/release-e2e-workflow-template.yml
peterzhuamazon: Got it.

PR #1587, opened 2024-10-16
[Maps plugin] Skip new home UI test if disabled in OSD

junqiu-lei:
### Description
Skip new home UI test if disabled in OSD

### Check List

- [x] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1708, opened 2025-02-12
[Maps Plugin] Fix disable check in Maps new home integ tests

q-andy:
### Description
Fix overly strict disable check in flaky integ test added in #1587

When an element is disabled, the disabled property is not always set to `disabled='disabled'`. Sometimes it is set with `disabled=''` or as a property without a value, `disabled`. In those cases, the test would not be skipped and the integ test fails since the switch is still disabled.

### Check List


ruanyl (approved):

heemin32 (approved):
LGTM. Thanks!

peterzhuamazon (approved):

peterzhuamazon:
Requested by @q-andy to backport to 2.17.

PR #1794, opened 2025-05-29
update the logic for hits check

angle943:
### Description

- This test has been failing a lot. See https://github.com/opensearch-project/OpenSearch-Dashboards/actions/runs/15307580055/job/43077057735?pr=9813 as an example.
- This updates the logic so that we are actually only checking the hit count goes back to the correct number as that is what the test is testing for.

### Check List

### DIFF
--- a/cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/data_explorer/discover.spec.js
+++ b/cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/data_explorer/discover.spec.js
@@ -133,14 +133,17 @@ describe('discover app', { scrollBehavior: false }, () => {
     });
 
     it('should reload the saved search with persisted query to show the initial hit count', function () {
+      // set current hit count as alias
+      cy.getElementByTestId('discoverQueryHits').invoke('text').as('hits');
       // apply query some changes
       cy.setTopNavQuery('test');
       cy.verifyHitCount('22');
 
       // reset to persisted state
       cy.getElementByTestId('resetSavedSearch').click();
-      const expectedHitCount = '14,004';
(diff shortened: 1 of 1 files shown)


cypress/integration/core-opensearch-dashboards/opensearch-dashboards/apps/data_explorer/discover.spec.js
```diff
@@ -133,14 +133,17 @@ describe('discover app', { scrollBehavior: false }, () => {
...
       cy.setTopNavQuery('test');
       cy.verifyHitCount('22');
 
       // reset to persisted state
       cy.getElementByTestId('resetSavedSearch').click();
-      const expectedHitCount = '14,004';
-      cy.verifyHitCount(expectedHitCount);
+      cy.get('@hits').then((hits) => {
+        cy.verifyHitCount(hits);
+      });
```
joshuali925, line 146: i don't have context on this test case, but would this just verify that the actual number is the same as itself?

joshuali925: actually i think i understand, we just want to make sure resetSavedSearch resets to the same number regardless of what it is

angle943: yeah. We first grab the original hit count and save it for the alias `@hits`.

Then we modify the query to get a different hit count (22).

Then we click `resetSavedSearch`, and then verify that the hit count is back to the original (by comparing it with the alias `@hits`

ashwin-pc (approved):

kavilla:
If we haven't backlogged an issue then we prolly should just to verify why the change occurred.

PR #2061, opened 2026-06-08
Fix flaky query-insights cypress tests on Windows

dzane17:
## Summary

Fix the persistent failures in the `queryInsightsDashboards` integ-test on Windows seen in builds [#8601](https://build.ci.opensearch.org/blue/organizations/jenkins/integ-test-opensearch-dashboards/detail/integ-test-opensearch-dashboards/8601/pipeline/) and [#8605](https://build.ci.opensearch.org/blue/organizations/jenkins/integ-test-opensearch-dashboards/detail/integ-test-opensearch-dashboards/8605/pipeline/) (autocut [#506](https://github.com/opensearch-project/query-insights-dashboards/issues/506)). The same tests pass on Linux ARM64 (build #8584) — the failures only reproduce on the Windows test agent, which is running cached Cypress 9.5.4 instead of the package-declared 13.17.0. The race conditions exposed by the slower Windows + older Cypress combination are real test-side bugs that this PR addresses with retry-friendly assertions.

### `1_top_queries_spec.js`

- **QUERY ONLY / GROUP ONLY header tests:** reorder `assertRowCountEquals(...)` BEFORE `getHeaders().should('deep.equal', ...)`. `assertRowCountEquals` uses `.should('have.length', N)` which retries until React state populates `items`; once row count matches, `columnsToShow` has evaluated against the populated data, so the deep-equal header assertion reads stable headers. Without this reorder, headers were read while `items.length === 0` → `defaultColumns` was rendered → header count mismatch.
- **Stats & Visualizations and DynamicSearchBar `beforeEach`:** the page fires three sequential GETs (`/api/top_queries/latency`, `/cpu`, `/memory`) and only flips `loading=false` after all three resolve. Wait for all three `@topQueries` plus the table to render so React state is settled and the conditionally-rendered `DynamicSearchBar` is stable before tests interact with it. This fixes the `'shows empty state when Group mode is selected'` DOM-detach failure and the 8 `cy.type` "disabled element" failures.

### `5_live_queries_spec.js`

- **`beforeEach`:** disable auto-refresh after the initial fetch so the inflight-queries table stops re-rendering every 5s. The auto-refresh `setInterval` was the root cause of the pagination "detached from DOM" error and the WLM-disabled "never did" content match.
- **Pagination test:** use the stable `[data-test-subj="pagination-button-1"]` selector (which appeared in the original failure log) instead of `.contains('2')`, and assert `aria-current="true"` to confirm the click took effect.
- **WLM-disabled-text test:** swap `cy.get('tbody tr').first().within(...)` for `cy.contains('tbody td', text)` which requeries the parent on every retry, surviving any incidental React reconciliation.
- **`'disables auto-refresh when toggled off'`:** since `beforeEach` now leaves auto-refresh off, this test re-enables it then toggles off so the observable transition to disabled is still exercised.

A companion PR with the equivalent change is open against the plugin repo to keep the cypress specs in sync. Plan to cherry-pick to `main` after merge.

## Test plan

- [ ] Re-run integ-test-opensearch-dashboards on a Windows distribution build with this branch and confirm queryInsightsDashboards passes on both with-security and without-security configs.
- [ ] Linux ARM64 / x64 integ-test still passes (no regressions).
- [ ] Existing `'disables auto-refresh when toggled off'` and the auto-refresh interval tests continue to behave correctly.

### DIFF
--- a/cypress/integration/plugins/query-insights-dashboards/1_top_queries_spec.js
+++ b/cypress/integration/plugins/query-insights-dashboards/1_top_queries_spec.js
@@ -390,8 +390,11 @@ describe('Query Insights — Dynamic Columns (QUERY ONLY fixture)', () => {
       'WLM Group',
       'Total Shards',
     ];
-    getHeaders().should('deep.equal', expected);
+    // assertRowCountEquals uses .should('have.length', N) which retries until items
+    // has populated; once row count matches, columnsToShow has evaluated against the
+    // populated data and headers are stable for the deep-equal assertion.
     assertRowCountEquals(getRowsFromRaw(QUERY_ONLY).length);
+    getHeaders().should('deep.equal', expected);
   });
 });
 
@@ -418,8 +421,11 @@ describe('Query Insights — Dynamic Columns (GROUP ONLY fixture)', () => {
       'Average CPU Time',
       'Average Memory Usage',
     ];
-    getHeaders().should('deep.equal', expected);
+    // assertRowCountEquals uses .should('have.length', N) which retries until items
+    // has populated; once row count matches, columnsToShow has evaluated against the
+    // populated data and headers are stable for the deep-equal assertion.
     assertRowCountEquals(getRowsFromRaw(GROUP_ONLY).length);
+    getHeaders().should('deep.equal', expected);
   });
 
   it('renders dash in status column for group rows', () => {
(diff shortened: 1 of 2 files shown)


yujin-emma (approved):
Thanks

PR #2068, opened 2026-06-22
fix CVE issues

yyfamazon:
### Description
  
  This PR addresses multiple security vulnerabilities in dependencies and hardens CI workflow actions by pinning them to specific commit SHAs.

  - CVE-2026-4800 (lodash): Added override/resolution to ^4.18.0 → resolves to 4.18.1
  - CVE-2026-32141 (flatted): Added override/resolution to ^3.4.0 → resolves to 3.4.2
  - GHSA-5c6j-r48x-rmvq (minimatch): Added override/resolution to ^5.1.9, replacing vulnerable 3.x which used brace-expansion@1
  - CVE-2025-69873 (ajv ReDoS with $data option): Added scoped resolutions (webpack/**/ajv, table/ajv, ajv-formats/**/ajv) to force ^8.20.0 for ajv 8.x consumers while keeping eslint on compatible ajv 6.15.0
  - CVE-2026-33672 (picomatch): Added override/resolution >=2.3.2 to mitigate method injection in POSIX character classes
  

### Issues Resolved

[List any issues this PR will resolve]

### Check List

- [ ] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #2000, opened 2026-05-05
Upgrade Cypress from 9 to 13 for FTRepo

yyfamazon:
### Description

Upgrade Cypress from 9 to 13 for FTRepo

baseline: https://github.com/opensearch-project/opensearch-dashboards-functional-test/pull/1895
I compare the ci test with baseline, and fix all tests that passed in baseline in this pr.

Issues Resolved
https://github.com/opensearch-project/opensearch-dashboards-functional-test/issues/1996

### Check List

- [ ] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #2073, opened 2026-06-25
Fix mocha version collision breaking Cypress reporters

TackAdam:
## Problem

Downstream plugin E2E runs (e.g. `dashboards-observability` FTR) started failing in the `Before` hook with:

```text
TypeError: test.slow is not a function
    at Runner.<anonymous> (.../Cypress/9.5.4/.../node_modules/mocha-7.0.1/lib/reporters/base.js:308:30)
    at Runner.emit (node:events:406:35)
    at Reporter.emit (.../packages/server/lib/reporter.js:362:46)
    ...
```

This fired before any spec assertions ran, so every job using this functional-test repo as a dependency was red.

## Root Cause

The regression was introduced in **#2068** ("fix CVE issues"), which migrated this repo from **yarn → npm**: it deleted `yarn.lock`, removed the `packageManager` field, and added a batch of `overrides`.

The break is a Mocha version collision:

- Cypress 13.17.0 runs the test suite with its own bundled `mocha@7.0.1` — note the stack trace path: `mocha-7.0.1/lib/reporters/base.js:308`.
- This repo wires up `cypress-multi-reporters` + `mocha-junit-reporter` (see `reporter-config.json`). Both declare `mocha` as a `peerDependency`, not a direct dependency.
- Under Yarn Classic (the previous setup), peer dependencies were not auto-installed, so no standalone `mocha` existed in `node_modules`. The reporters simply bound to Cypress's bundled `mocha@7.0.1`, and everything matched.
- Under npm v7+ (the new setup after #2068), peer dependencies are auto-installed. npm hoisted a top-level `mocha@11.7.6` to satisfy the reporters' peer ranges. At runtime, Cypress's bundled `mocha@7.0.1` reporter `base.js` received `Runner`/`Test` objects from `mocha@11`, where `test.slow` is no longer on the `Test` prototype the v7 reporter expects → `test.slow is not a function` at `base.js:308`.

In short: nothing in the test code changed — the dependency-manager switch silently introduced a second, incompatible Mocha into the tree.

## Fix

**Pin Mocha to Cypress's bundled version** via an override so both reporters resolve against the same Mocha that Cypress actually runs:

```json
"overrides": {
  ...
  "ajv": "^8.20.0",
  "mocha": "7.0.1",
  ...
}
```

The `package-lock.json` is updated **incrementally from the working `main` lockfile** (applying only the new `mocha` override), rather than regenerated from scratch. This matters: a from-scratch `npm install` on some npm versions re-resolves the existing `ajv` override down to `6.15.0` at the top level, which then puts the lockfile out of sync with the `ajv: ^8.20.0` override and breaks `npm ci`. Building incrementally preserves top-level `ajv@8.20.0` (in sync) while keeping `@eslint/eslintrc`'s nested `ajv@6.15.0`.

This single change resolves three failure modes that all traced back to #2068's overrides:

1. **Cypress reporters** — `mocha@7.0.1` is now the only Mocha in the tree, so `test.slow` exists.
2. **All E2E ciGroups** — they run `npm ci` first; the in-sync lockfile fixes the `npm error ... Missing: ajv@8.20.0 from lock file` (EUSAGE) that was failing every group before any test ran.
3. **Lint Checker** — the in-sync lockfile keeps `@eslint/eslintrc` on `ajv@6.15.0`, fixing `TypeError: Cannot set properties of undefined (setting 'defaultMeta')`.

### Also included

- **Pin `dorny/paths-filter` to a commit SHA** in `dashboard-variable-release-e2e-workflow.yml`. This workflow was the only one still on the unpinned `@v2` (missed in c791b5b "Pin GitHub Actions to commit SHAs"), so the org's SHA-pinning policy was failing its `changes` job. Uses the same SHA (`4512585405083f25c027a35db413c2b3b9006d50 # v2`) already used by the other 20 e2e workflows.

## Verification

All verified locally before push:

- `npm ls mocha` shows a single resolved version — both reporters use `mocha@7.0.1` (overridden):

```text
├─┬ cypress-multi-reporters@1.6.4
│ └── mocha@7.0.1 overridden
└─┬ mocha-junit-reporter@2.2.1
  └── mocha@7.0.1 deduped
```

- `Test.prototype.slow` is confirmed to be a function on `mocha@7.0.1` (the exact call that was throwing).
- Both `mocha-junit-reporter` and `cypress-multi-reporters` `require()` cleanly against `mocha@7.0.1`.
- `npm ci` succeeds (lockfile in sync — no EUSAGE).
- `eslint . --ext .js` passes.

## Notes / Follow-ups

- This change is intentionally scoped to the test break and the missed action pin.
- The other overrides added in **#2068** (`lodash ^4.18.0`, `minimatch ^5.1.9`, etc.) are left untouched — they resolve and install fine. A separate pass to sanity-check those CVE pins may be worthwhile.

### Check List


peterzhuamazon:
Hi @TackAdam there is a reason we used to have two lock.
I am not sure what was the intent from #2068.

But:
1. package-lock.json is specifically used for npm install in integtest.sh
2. yarn.lock is specifically used for advisories scan + yarn start commands.

~~A good migration would be completely remove npm and use yarn exclusively, and remove npm.~~

~~That means:~~
~~1. remove package-lock.json~~
~~2. Add back yarn.lock~~
~~3. update integtest.sh and replace `npm install` with `yarn`.~~

In close observation we have to keep both.
Npm install also ensures that cypress custom build cache is used as-is.
Which is not in conflict with yarn install.

And yarn.lock is required for advisories to take effect.

We should revert the behavior of the lock files in #2068 and keep both.

Thanks.

cc: @yyfamazon @wanglam @yubonluo

SuZhou-Joe:
@peterzhuamazon The yarn.lock and packageManager was accidentally introduced by #2000, and it is not practical to keep 2 lock files in a repo, introducing issues like above and inconsistent lock versions.  Do we have any plan to make `advisories` support package-lock.json? And seems package-lock.json is already supported by advisories:
image

SuZhou-Joe:
We can merge this fix anyway, the yarn revert can be done in another separate PR.

SuZhou-Joe (approved):

peterzhuamazon:
After discussion, we can merge this as is and not introduce yarn.lock anymore as it was introduced in #2000 incorrectly.

in other repo there used to be scanning errors when both lock file present, but since ftrepo never fail the scan makes I think yarn lock was presented from the start.

We can merge this as is and monitor Jenkins. Thanks.

PR #2102, opened 2026-07-18
Stabilize flaky search-relevance dashboards e2e tests (judgment async, page load, combo options)

will-hwang:
### Description

Fixes intermittent failures in the search-relevance-dashboards Cypress specs (`5_judgment_creation` and `6_experiment_creation`), which were flaky against a managed OpenSearch domain while passing when run manually. Three related root causes, each fixed:

1. **Async judgment ratings race** — UBI/LLM judgments are created asynchronously: the create call returns `200` immediately with `status=PROCESSING` and empty ratings, then ratings compute in the background and status flips to `COMPLETED`. The detail view fetches once on mount and does not poll, so asserting right after creation raced the background job and intermittently rendered an empty ratings table. Added a `waitForJudgmentCompleted()` helper that polls the judgment listing until `COMPLETED` before navigating to the detail view / consuming the judgment. Used in spec 5 and in spec 6's setup.

2. **`cy.waitForLoader()` hang on the search evaluation page** — this page's on-mount data fetches can keep the browser `load` event pending, so `waitForLoader()`'s `homeIcon` assertion timed out even though the page was usable. Replaced it with a wait for the first combo box to render and become enabled.

3. **Combo options stuck on "Loading..."** — the query set / search configuration / judgment combo boxes fetch their options only on mount. Occasionally a combo stayed stuck on "Loading..." (its on-mount fetch didn't populate). Since waiting longer can't help (options are only fetched on mount), wrapped the form interaction in a bounded retry (up to 3): open the query set combo as a canary and, if its option hasn't rendered, re-visit the route to re-fire the fetches before retrying. Re-mount is done via a fresh `cy.visit()`, **not** `cy.reload()` — reloading this client-side hash route lands on a blank OpenSearch Dashboards app shell, whereas a fresh visit re-bootstraps the SPA like the initial load.

### Issues Resolved

N/A

### Check List
- [x] All tests pass (search-relevance specs run locally against OpenSearch + OpenSearch-Dashboards with the search-relevance, ubi, and ml plugins)
- [x] New functionality has been documented via inline comments explaining the non-obvious timing/SPA behavior

### DIFF
--- a/cypress/integration/plugins/search-relevance-dashboards/5_judgment_creation.spec.js
+++ b/cypress/integration/plugins/search-relevance-dashboards/5_judgment_creation.spec.js
@@ -10,6 +10,7 @@ import { BASE_PATH } from '../../../utils/base_constants';
 import {
   initializeUbiIndices,
   enableWorkbenchUI,
+  waitForJudgmentCompleted,
 } from '../../../utils/plugins/search-relevance-dashboards/common-setup';
 
 const PLUGIN_TIMEOUT = 30000;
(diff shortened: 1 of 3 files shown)


will-hwang:
@maintainers Could a maintainer please add the **`backport 3.7`** label? This fixes an intermittent failure currently affecting the `Search Apps E2E OS` step on the 3.7 release line (the same bug exists on `main` and `3.7` — identical files). I don't have permission to apply labels as an external contributor. Thanks!

SuZhou-Joe (approved):

